PR #28, opened 2018-02-06
fix role and gcloud instance cleanup

rphillips:
adds required serviceAccountActor role and quiets the instance deletion.


nof20:
Thanks @rphillips !

PR #30, opened 2018-02-20
Make builder

rvkubiak:
Allows execution of simple Make targets such as targets that execute shell commands to prepare a workspace.  This can be useful when replacing existing Make builds with Google Container Builder configs.

### DIFF
--- a/make/cloudbuild.yaml
+++ b/make/cloudbuild.yaml
@@ -1,0 +1,11 @@
+# In this directory, run the following command to build this builder.
+# $ gcloud container builds submit . --config=cloudbuild.yaml
+
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/make', '.']
+- name: 'gcr.io/$PROJECT_ID/make'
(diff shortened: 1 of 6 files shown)


Philmod:
Could you add an `example` directory ?

make/cloudbuild.yaml
```diff
@@ -0,0 +1,11 @@
+# In this directory, run the following command to build this builder.
+# $ gcloud container builds submit . --config=cloudbuild.yaml
+
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/make', '-t', 'gcr.io/$PROJECT_ID/make', '.']
```
ImJasonH: This line tags the image twice with the same tag, is that intentional?

rvkubiak:
Updated to squash commits.
Change entrypoint from bash to make to be consistent with other builders.
Added examples directory.
Removed duplicate tagging in Docker build.

make/cloudbuild.yaml
rvkubiak: Not intentional, fixed.

PR #31, opened 2018-02-21
Add hub builder

ImJasonH:


nof20:
LGTM.  No example because we'd need a Github account.

PR #32, opened 2018-02-22
Upgrade helm to 2.8.1, support regional GKE clusters, cleanup

victortrac:
Changes:
* Upgrade helm to 2.8.1
* Cleaned up Dockerfile to be more readable
* Updated /helm/cloudbuilder.yaml to support pushing to repos outside of GCP project that built the image (to push to a public repo, for example)
* Pull in GCP/GKE projects, GKE cluster, region, & zone from env vars in the helm docker image if supplied. Otherwise, use the same gcloud commnads to discover.
* Default to beta region gcloud API if a region is passed in

### DIFF
--- a/helm/Dockerfile
+++ b/helm/Dockerfile
@@ -1,9 +1,21 @@
 FROM gcr.io/cloud-builders/gcloud
+
+ENV VERSION v2.8.1
+
 COPY helm.bash /builder/helm.bash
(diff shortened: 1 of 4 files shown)


rimusz:
LGTM 👍

nof20:
Thanks so much!

PR #33, opened 2018-02-23
Docker image that extends the golang image to add godep support

rvkubiak:
Godep is an official golang tool for dependency management.  It is
heavily used in the Kubernetes project, but is not included in the base
go build images.

### DIFF
--- a/godep/Dockerfile
+++ b/godep/Dockerfile
@@ -1,0 +1,12 @@
+FROM golang
+
+ENV GOPATH /gopath/
+ENV PATH $GOPATH/bin:$PATH
+
+RUN go version
+RUN go get github.com/tools/godep
+RUN godep version
+RUN apt-get update && \
+    apt-get install -y build-essential
+
+ENTRYPOINT ["/gopath/bin/godep"]
--- a/godep/README.md
+++ b/godep/README.md
@@ -1,0 +1,17 @@
+# Godep
+
+This build step provides a container based on the golang container that adds in
+the [Godep](https://github.com/tools/godep) tool.  Godep is an official golang
+tool for dependency management.
+
+This image uses godep as the entrypoint and takes in go commands as arguments.
+An example use case for Google Container Builder:
+
+```
+- name: gcr.io/cloud-builders-community/godep
+  dir: "/workspace/src/github.com/GoogleCloudPlatform/k8s-stackdriver/event-exporter"
+  args: ["go", "test", "./..."]
+```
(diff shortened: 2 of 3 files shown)


godep/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM golang
```
Philmod: Would it make sense to extends the Go builder we have? https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/go

godep/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+ENV GOPATH /gopath/
+ENV PATH $GOPATH/bin:$PATH
+
+RUN go version
+RUN go get github.com/tools/godep
+RUN godep version
```
Philmod: you can remove this probably

godep/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+This build step provides a container based on the golang container that adds in
+the [Godep](https://github.com/tools/godep) tool.  Godep is an official golang
+tool for dependency management.
+
+This image uses godep as the entrypoint and takes in go commands as arguments.
+An example use case for Google Container Builder:
+
+```
+- name: gcr.io/cloud-builders-community/godep
```
Philmod: Could you create an `example` directory with a full working example?

ImJasonH:
The GitHub page for `godep` says "Please use dep or another tool instead." so it sounds like this tool isn't supported by the developer anymore.

Does this indicate we should recommend using some other dep tool instead?

rvkubiak:
I believe 'go dep' is the replacement for 'godep go', however a significant number of Kubernetes images still use godep.  I have since learned that running 'godep go build' is really only necessary for go1.4 and earlier.  Since go1.5 godep uses the vendor subdir for storing dependencies which is included in the default search path.

As a result, the 'godep go build' commands that are prevalent throughout Kubernetes makefiles aren't actually necessary, and simply running 'go build' should work in nearly every case.

I'll probably maintain a version of this builder for Kubernetes to maintain backwards compatibility, but since go1.4 is no longer supported I doubt anyone else will have any use for this.

ImJasonH:
@rvkubiak Does your comment mean that you no longer need this PR?

rvkubiak:
Yes, I think this PR is not valuable.  I'll maintain a copy for Kubernetes, but there are better ways to handle go dependencies today.

PR #36, opened 2018-03-01
Test PR integration with GCB - Do not submit !

Philmod:


cloudbuild.yaml
```diff
@@ -6,11 +6,33 @@ steps:
...
+          echo "$d failed" >> ${failure_file}
+        fi
+      ) &
     done
 
+    wait
+
+    # Check if there is any failure.
+    num_failures=$(wc -l < "${failure_file}")
+    if [[ $num_failures > 0 ]]; then
```
ImJasonH: `if [[ -s ${failure_file} ]]; then`

cloudbuild.yaml
```diff
@@ -6,11 +6,33 @@ steps:
   args:
   - '-c'
   - |
+    # File that contains failures.
+    failure_file=failure.log
+    touch ${failure_file}
```
ImJasonH: You don't have to `touch` this, I don't think.

cloudbuild.yaml
```diff
@@ -6,11 +6,33 @@ steps:
...
+      echo "Some builds failed"
+      cat ${failure_file}
+      echo "Exiting."
+      exit 1
+    fi
+
+    echo "All builds succeeded."
+
+
 timeout: 1200s
```
ImJasonH: Might need a longer `timeout` since we'll be kicking off N concurrent builds, where N>10 will be queued until others complete.

Issue #40, opened 2018-03-02
remote-builder doesn't have a cloudbuild.yaml


Philmod:
Neither an example directory.

nof20:
Fixed!  See https://github.com/GoogleCloudPlatform/cloud-builders-community/pull/44

PR #41, opened 2018-03-04
Pull Request integration

Philmod:
**tl;dr** Using GCF, the PRs code is tested on GCB, and the status reported back on the Github PR.

The process works like this:
- Github sends a message to a GCF function (webhook) when a PR is created/modified
- GCF creates a first build that fetches the PR code then creates as many sub-builds as directories. The main build carries some information in the build tags: repo, PR number, sha.
- Another GCF function ingests builds result from Pubsub, and only processes the builds that contains the expected tags. The function then reports the build's status to the Github Pull Request, after getting Github token using Google Cloud KMS.

This code only contains the main GCB build configuration. The code of the 2 GCF functions are not copied here.

process

The reason the status is red is https://github.com/GoogleCloudPlatform/cloud-builders-community/issues/40. All the others builders build.


Philmod:
Done.

More failures now: 
```
Step #3: remote-builder/ failed
Step #3: kubectl_wait_for_job/examples/pods-list failed
Step #3: helm/examples/pods-list failed
Step #3: packer/examples/gce failed
```

nof20:
@ImJasonH let's talk about that.  Another solution might be to make the examples hermetic, and e.g. for GKE spin up a cluster before they run the example command.  This gives a better user experience (lower barrier to entry) and allows the examples to act like tests of the builders.

ImJasonH:
Sure, that would be nice. But I'd like to get this change in by itself
before we fix up all the broken examples, which might take a bit of work.

One disadvantage of requiring examples to be hermetic tests of the
builder's functionality is that it raises the bar for contributing even
higher, and introduces more room for breakages that contributors (or our
team) have  to investigate and fix.

nof20:
Good point.  SGTM

PR #42, opened 2018-03-09
Add skaffold step (alpha)

ahmetb:
Derived from the official kubectl builder, this works with GKE as well.

Explicitly marking this as `:alpha` so users are aware what they are deploying
is pre-stable material.

I'm using this right now and seems to be working fine.

### DIFF
--- a/skaffold/README.md
+++ b/skaffold/README.md
@@ -1,0 +1,34 @@
+# `gcr.io/cloud-builders-community/skaffold:alpha`
+
+This Container Builder build step runs
+[`skaffold`](https://github.com/GoogleCloudPlatform/skaffold/) for Google
+Kubernetes Engine (GKE) clusters.
+
+> **NOTE:** This builder image is using pre-stable versions of Skaffold and
+> is **NOT RECOMMENDED** for production use.
+
+## Using this builder
+
+This builder is derived from the [`kubectl` builder](https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/kubectl)
+and therefore supports the same environment variables. Read the `kubectl` builder
+documentation to use Skaffold builder.
+
+## Example: Deploying to Google Kubernetes Engine
+
+To run `skaffold run` in your cloudbuild.yaml, include this step with the
+correct Kubernetes Engine cluster name and compute zone:
+
+```
+steps:
+- name: gcr.io/cloud-builders-community/skaffold:alpha
+  args: ['run', '-f=skaffold.yaml']
+  env:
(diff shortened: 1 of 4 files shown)


ahmetb:
We should ideally graduate this to official builders, but this does the job for now.
/cc @viglesiasce @r2d4

ahmetb:
PTAL @bendory @ImJasonH

skaffold/skaffold.bash
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+# Invoke kubectl.bash to process CLOUDSDK_* environment variables and make a
```
ImJasonH: Is it worth somehow modifying `kubectl` to make this easier, without having to pollute logs with `version`? If anything comes to mind, feel free to submit an issue or PR on the official builder repo.

skaffold/cloudbuild.yaml
```diff
@@ -0,0 +1,5 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/$PROJECT_ID/skaffold:alpha', '.']
+
```
ImJasonH: Can you add a simple sanity test here that you can invoke the builder at all? Even something as simple as `skaffold version`.

skaffold/skaffold.bash
ahmetb: Imo this is probably necessary for two reasons:

1. `kubectl version` verifies that it can actually talk to the GKE cluster. We can always discard the output `2>&1 1>/dev/null` but that wouldn't work when it actually failed.

skaffold/cloudbuild.yaml
ahmetb: @ImJasonH I'm afraid this will be difficult to do here. Because the startup script of this image will call `kubectl.bash` which will require a real GKE cluster (and CLOUDSDK_CONTAINER_CLUSTER, CLOUDSDK_COMPUTE_ZONE set). So it doesn't seem doable here.

I can however add a `RUN skaffold version` to the Dockerfile, which can work?

ahmetb:
@ImJasonH can you please take a look if my image name in README.md title is wrong?

skaffold/README.md
```diff
@@ -0,0 +1,34 @@
+# `gcr.io/cloud-builders-community/skaffold:alpha`
```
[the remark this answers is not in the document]
ahmetb: Darn, I thought I understood the community builders proposals. :) Apparently not.

I'll just send a patch to correct this part.

PR #44, opened 2018-03-15
Fix several bugs with Remote Builder

nof20:
Fixes several bugs/issues with Remote Builder:

- Moves to "builder builder" pattern by adding a root `cloudbuild.yaml` and a separate one for the canonical example
- Fixes lack of compute API permission in new projects
- Fixes typo in IAM role
- Adds visible output from example script
- Specifies a zone (`us-central1-f`) in canonical example


remote-builder/examples/cloudbuild.yaml
```diff
@@ -0,0 +1,21 @@
+steps:
+- name: gcr.io/$PROJECT_ID/remote-builder
+  waitFor: ["-"]
+  env:
+    # Use Container Optimized OS
+    # https://cloud.google.com/container-optimized-os/docs/
+    - INSTANCE_ARGS=--image-project cos-cloud --image-family cos-stable
+    - ZONE=us-central1-f
+    - USERNAME=cloud-user
```
Philmod: Where does `cloud-user` come from?

remote-builder/examples/test-scripts/no-op.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -xe
+echo Hello World from Remote Builder!
+true
```
Philmod: The last line is not needed, the second will return a 0 exit status already.

remote-builder/examples/cloudbuild.yaml
viglesiasce: This is the default user for the COS images

remote-builder/README.md
```diff
@@ -60,58 +74,7 @@ build step in the `env` parameter:
...
-```
-
-4. You should now see 2 instances being provisioned in parallel then the `test/no-op.sh` being
-run inside containers based on the `ubuntu:16.04` Docker image.
+To give it a try, see the `examples` directory.
 
-## Trade-offs
+This is not an official Google product.
 
-1. Paying for builder + VM
```
viglesiasce: The trade offs were removed. Was that intentional?

remote-builder/examples/README.md
```diff
@@ -0,0 +1,19 @@
...
+## Quick Start
+
+In this simple example, you will run a script inside of containers on two instances in
+parallel. You will use the Container Optimized OS image to provide an image with Docker
+pre-installed. The build request runs the `test/no-op.sh` script from this directory.
+
+To run, execute this command:
+
+```shell
+gcloud container builds submit --config cloudbuild.yaml .
```
viglesiasce: For the quick start do we need to point them at actually creating the builder otherwise their build will fail.

remote-builder/examples/README.md
```diff
@@ -0,0 +1,19 @@
+# Remote Builder Examples
```
viglesiasce: The folder is called `examples` but is currently set to only have 1 example. Should we just rename to `example`?

remote-builder/README.md
nof20: Whoops, no!  Copy & paste, added back.

remote-builder/examples/README.md
nof20: Good point, added

remote-builder/examples/README.md
nof20: Renamed the title in the README but let's leave the directory named examples, since there are some other `cloudbuild.yaml` files in there, if that's ok?

remote-builder/examples/test-scripts/no-op.sh
nof20: Removed, thx

remote-builder/README.md
```diff
@@ -60,58 +74,12 @@ build step in the `env` parameter:
...
 | INSTANCE_ARGS| Parameters to the instance creation command. For a full list run `gcloud compute instances create --help`| `--preemptible` |
 
-## Quick Start
-
-In the following example, you will run a script inside of containers on two instances in
-parallel. You will use the Container Optimized OS image to provide an image with Docker
-pre-installed. The build request runs the `test/no-op.sh` script from this directory.
-
-1. Add the Cloud Build service account as a serviceAccountUser and instanceAdmin:
+To give it a try, see the `examples` directory.
```
viglesiasce: You should be able to make this a link...

Something like:

To give it a try, see the [examples](examples) directory.

nof20: 👍Done

PR #45, opened 2018-03-20
Upgrade to helm 2.8.2

chainlink:


chainlink:
Hey, I can't see why this failed, happy to fix if you post the logs though!

ImJasonH:
It seems to be an error with the helm builder before your change. Substituting `$PROJECT_ID` for `$_PROJECT_ID` doesn't cascade; the `$PROJECT_ID` is not further substituted for the build's project ID, it's just the string `$PROJECT_ID`, which is invalid in an image name.

Are you able to build this change locally, using `gcloud container builds submit`? It looks like there are other blocking errors besides the `$_PROJECT_ID` substitution.

I have no problem merging your change since it seems unrelated, but somebody should take some time to fix up the helm builder so it builds at all.

PR #46, opened 2018-03-20
Fix helm builder

ImJasonH:
- substituting `$_PROJECT_ID` with `$PROJECT_ID` doesn't cascade; `$PROJECT_ID` is not further substituted with the build's project ID
- the `docker build` should build the contents of `.`, not the `helm/` directory, which is not found inside `helm/`


nof20:
LGTM.

PR #50, opened 2018-04-12
Add bq builder

tswast:
This builder is very similar to the [gsutil
builder](https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/gsutil)
with the one addition that the Dockerfile must create an empty
`~/.bigqueryrc` file so that the `bq` command knows to inherit
credentials and configuration from `gcloud`.


bq/README.md
```diff
@@ -0,0 +1,44 @@
+# gsutil
```
ImJasonH: This README is still all about gsutil.

Philmod:
```
Step #2: Already have image: gcr.io/community-builders/bq
Step #2: BigQuery error in query operation: Access Denied: Project community-builders:
Step #2: The user X@cloudbuild.gserviceaccount.com does not have
Step #2: bigquery.jobs.create permission in project community-builders.
Finished Step #2
ERROR
```

I guess we would have to setup our GCP project

tswast:
@Philmod correct. I added instructions in the example's README.

> ... assign the [BigQuery Data
Editor](https://cloud.google.com/bigquery/docs/access-control#permissions_and_roles)
role to [your Container Builder service
account](https://cloud.google.com/container-builder/docs/securing-builds/set-service-account-permissions)

To just run the builder you don't need full editor, though, just BigQuery User will be fine to run queries against public datasets.

bq/README.md
tswast: Oops. Fixed, thanks.

tswast:
@Philmod I have also added a note to the `cloudbuild.yaml` file.

tswast:
Ping? I made the requested changes months ago.

bq/examples/bq_load/cloudbuild.yaml
```diff
@@ -0,0 +1,8 @@
+steps:
+# Download a sample data file (could have been generated by a binary).
+- name: 'gcr.io/cloud-builders/wget'
```
ImJasonH: Might be simpler as `gsutil cp` but I won't fight for it.

tswast: The bucket is public, so either should work fine.

PR #51, opened 2018-04-14
Composer Package Manager

diogotozzi:
Added support to Composer: PHP most used Package Manager

### DIFF
--- a/composer/Dockerfile
+++ b/composer/Dockerfile
@@ -1,0 +1,1 @@
+FROM composer
--- a/composer/Dockerfile
+++ b/composer/Dockerfile
@@ -1,1 +1,3 @@
 FROM composer
+
+ENTRYPOINT ["composer"]
--- a/composer/README.md
+++ b/composer/README.md
@@ -1,0 +1,14 @@
+# Composer
+
+[Composer](https://getcomposer.org/) is a popular PHP open-source package manager.
+
+## Usage:
+
+```
+steps:
+- name: 'gcr.io/cloud-community-builders/composer
+  args:
+    - '-c'
+    - |
+      composer install
+```
--- a/composer/README.md
+++ b/composer/README.md
@@ -6,9 +6,6 @@
 
 ```
 steps:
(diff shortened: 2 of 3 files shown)


diogotozzi:
I signed it!

composer/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM composer
```
ImJasonH: The convention used in other builders is to set the `ENTRYPOINT` as well, in this case to `composer`

This allows a build step to invoke it with just args:

```
- name: gcr.io/foo/composer
  args: ['install']
```

composer/cloudbuild.yaml
```diff
@@ -0,0 +1,5 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/$PROJECT_ID/composer', '.']
+
```
ImJasonH: Can you add a simple sanity test that ensures you can invoke `composer install` using the newly built builder?

composer/Dockerfile
diogotozzi: Done!

diogotozzi: Done

composer/README.md
```diff
@@ -0,0 +1,14 @@
+# Composer
+
+[Composer](https://getcomposer.org/) is a popular PHP open-source package manager.
+
+## Usage:
+
+```
+steps:
+- name: 'gcr.io/cloud-community-builders/composer
+  args:
```
ImJasonH: This example is out of date now, should be `args: ['install']`

composer/README.md
```diff
@@ -0,0 +1,14 @@
+# Composer
+
+[Composer](https://getcomposer.org/) is a popular PHP open-source package manager.
+
+## Usage:
+
+```
+steps:
+- name: 'gcr.io/cloud-community-builders/composer
```
ImJasonH: This image name should be `gcr.io/$PROJECT_ID/composer` since we expect/require users to first build the image in their own project.

PR #52, opened 2018-04-15
Add cron-builder

nof20:
Add cron-builder.


cron-builder/README.md
```diff
@@ -0,0 +1,61 @@
...
+export PROJECT=$(gcloud info --format='value(config.project)')
+gsutil mb gs://$PROJECT-source
+tar cvzf source.tar.gz [your files here]
+gsutil cp source.tar.gz gs://$PROJECT-source
+```
+
+Then, to reference it in your `cloudbuild.yaml`, add the following lines:
+
+```
+source:
```
ImJasonH: You can't specify source in your cloudbuild.yaml, you have to specify it on the command line using `gcloud` or in a triggered build the source is the source that changed.

ImJasonH: Oh I see; you're allowing users of this Cron service to specify source in the YAML. That might become confusing since we don't normally allow that. The Cron cloudbuild.yaml won't be valid if executed directly.

nof20: Oh, I see - it only works because I parse the YAML and submit through the Go API.

I'm trying to make it as easy as possible for folks to use.  Can you think of a better way of doing it?

nof20:
Ping @ImJasonH - let me know if my changes are OK.

PR #53, opened 2018-04-23
Initial revision of gcs-cache builder.

bendory:


ImJasonH:
Can this leverage `gcs-fetcher` to benefit from incremental upload?

I have a TODO to package manifest-based upload as a builder image which is basically all you'd need. Then `gcs-cache` can just be documentation about how to use `gcs-fetcher` and `gcs-uploader`.

bendory:
@ImJasonH Nice idea, and likely gives a better speed-up than this zip + ship implementation. Would the "API" (==args) to `pull` and `push` commands be the same as current? If so we can merge this as is and then modify when your TODO is completed.

bendory:
Closing this PR -- interested users should use https://github.com/GoogleCloudPlatform/cloud-builders/tree/master/gcs-fetcher instead.

PR #57, opened 2018-05-08
Add buildah builder

ImJasonH:
This builder is capable of building and pushing container images from Dockerfiles (using `buildah bud`) or using Buildah's own declarative style, without access to the Docker daemon socket.

### DIFF
--- a/buildah/Dockerfile
+++ b/buildah/Dockerfile
@@ -1,0 +1,31 @@
+FROM launcher.gcr.io/google/ubuntu16_04
+
+RUN \
+  # Install runc
+  apt-get update && \
+  apt-get -y install wget && \
+  wget -O /bin/runc https://github.com/opencontainers/runc/releases/download/v1.0.0-rc5/runc.amd64 && \
+  chmod +x /bin/runc && \
+  PATH=/usr/local/bin/:$PATH && \
+  runc --help && \
+  # Install buildah dependencies
+  apt-get -y install software-properties-common && \
+  add-apt-repository -y ppa:alexlarsson/flatpak && \
+  add-apt-repository -y ppa:gophers/archive && \
+  apt-add-repository -y ppa:projectatomic/ppa && \
+  apt-get -y -qq update && \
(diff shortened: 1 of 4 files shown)


buildah/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM launcher.gcr.io/google/ubuntu16_04
+
+RUN \
+  # Install runc
+  apt-get update && \
+  apt-get -y install wget && \
+  wget -O /bin/runc https://github.com/opencontainers/runc/releases/download/v1.0.0-rc5/runc.amd64 && \
+  chmod +x /bin/runc && \
+  PATH=/usr/local/bin/:$PATH && \
+  runc --help && \
```
bendory: should we delete this debug line?

buildah/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+  # Install buildah dependencies
+  apt-get -y install software-properties-common && \
+  add-apt-repository -y ppa:alexlarsson/flatpak && \
```
bendory: Can move these repositories up above line 5 and thereby get away with just one `apt-get update` call.

buildah/Dockerfile
ImJasonH: It's not just a debug line, it'll fail the build faster if runc isn't installed correctly, before going through all the rest.

buildah/Dockerfile
ImJasonH: I wanted to separate the build into "install runc" and "install buildah", and fail faster if runc didn't install correctly. I don't think it probably adds much latency to have these split, and I think the conceptual separation is worth the small amount of duplication.

jonjohnsonjr: Add a comment to that effect?

buildah/Dockerfile
jonjohnsonjr: Add a comment to that effect?

ImJasonH: Done (above)

nof20:
Merging approved PR.

PR #58, opened 2018-05-09
Fix skaffold releases url and docker config

stevenbarragan:
This ensures curl is installed before using it and cleans it up after.

It fixes the download url, since the current one doesn't points to the latest release.

It runs `gcloud auth configure-docker` and it sets the `DOCKER_CONFIG` since it default's to home but the last commands puts it on `/builder/home/.docker`, to let skaffold authenticated docker commands.

Not included in this PR but skaffold depends on `helm` to be installed as well, the appropriate thing to do would be to start FROM `gcr.io/cloud-builders/helm` but since its not an official builder yet what's the best approach here?

### DIFF
--- a/skaffold/Dockerfile
+++ b/skaffold/Dockerfile
@@ -1,9 +1,17 @@
 FROM gcr.io/cloud-builders/kubectl
 
 RUN mkdir -p /builder/bin && \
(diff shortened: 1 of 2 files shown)


nof20:
Thanks a lot for this!  Changes look good.

It's possible to use Helm with Skaffold, but it's not a mandatory dependency (see the list of deps in their [README](https://github.com/GoogleContainerTools/skaffold/blob/master/README.adoc)).  I think we can leave this optional for now, what do you think?

PR #63, opened 2018-05-25
Only print helm arguments when debug is true

defrank:
By default, all arguments passed to helm are printed to stdout.  This is an issue when using `--set-string` to override a secret value.  The value is echoed to the logs of CI/CD.

This PR is meant to disable this behavior by default.  Only enabling it if running with an environment variable of `DEBUG=true`.

### DIFF
--- a/helm/helm.bash
+++ b/helm/helm.bash
@@ -44,5 +44,7 @@ fi
 echo "Running: helm repo update"
 helm repo update
 
-echo "Running: helm $@"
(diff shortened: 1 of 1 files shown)


defrank:
I signed it!

Note that I use my personal email in GitHub, but signed with my work email.

defrank:
I signed it!

defrank:
I fixed my email!

defrank:
I signed it!

PR #65, opened 2018-06-03
add firebase cloudbuilder

petomalina:
Hi there,

I just started using cloud builder as a CI&CD for Firebase. The builder installs `firebase-tools` locally via npm to avoid permission errors from node-gyp. The `firebase.bash` script then wraps every call to provide firebase CI token.

Cheers,
Peter


petomalina:
@googlebot I signed it!

PR #66, opened 2018-06-06
remote-builder: Fix default username in README

cceckman:
If I'm reading [1] correctly, it's "admin" rather than "cloud-user".

[1] https://github.com/GoogleCloudPlatform/cloud-builders-community/blob/cab7b7b2d290c1ec461e50c2723265a4c6f3b1a8/remote-builder/run-builder.sh#L11


nof20:
Looks good to me, thanks a lot.

PR #68, opened 2018-06-08
update packer version to 1.2.4

Thomas-Roper:
Parameterize packer version and checksum as cloudbuild substitutions
Add defaults for packer version to 1.2.4


bendory:
Thanks for the contribution! LGTM.

PR #64, opened 2018-05-31
[dep] Add dep cloud builder

EngelVoelkersTECHNOLOGY:
This PR will add the dep cloud-builder as described in https://github.com/GoogleCloudPlatform/cloud-builders-community/issues/60.
This cloud builder is based on the official go cloud-builder and provides a Debian and Alpine image.

Issue #60, opened 2018-05-11
Add golang dep builder


ImJasonH:
Moved from https://github.com/GoogleCloudPlatform/cloud-builders/issues/275

christian-roggia:
I will open a PR this week about the community cloud builder for golang/dep.

PR #69, opened 2018-06-13
[dep] Add dep cloud builder

christian-roggia:
This PR will add the dep cloud-builder as described in #60.
This cloud builder is based on the official go cloud-builder and provides a Debian and Alpine image.

This PR is a duplicate of #64 aimed to solve the CLA issue. The previous PR will be closed.


dep/cloudbuild.yaml
```diff
@@ -0,0 +1,23 @@
...
+  args: ['build', '-f', 'Dockerfile.debian', '--tag=gcr.io/$PROJECT_ID/dep:debian', '.']
+
+# The alpine version is our default.
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['tag', 'gcr.io/$PROJECT_ID/dep:alpine', 'gcr.io/$PROJECT_ID/dep']
+
+# Run version to make sure that everything is working properly.
+- name: 'gcr.io/$PROJECT_ID/dep'
+  args: ['version']
+  env: ['GOPATH=.']
```
ImJasonH: Is it necessary for users to set this env to use this builder correctly? Can we set it for them by default? Ideally users wouldn't have to remember to add this, since it's fairly easy to forget.

dep/dep.ash
```diff
@@ -0,0 +1,19 @@
+#!/bin/ash
```
ImJasonH: Unless we expect a lot of opinionated usage, I think we should produce only one image instead of both Alpine and Debian. I think it was a mistake to support both flavors for `gcr.io/cloud-builders/go`, and over time I suspect we'll try to get rid of `go:alpine` altogether.

So my preference is that we only produce one `dep` builder, and that it's based on Debian.

WDYT? Do you need the Alpine image?

dep/Dockerfile.debian
```diff
@@ -0,0 +1,25 @@
+FROM golang
```
ImJasonH: Is it possible to base this on `gcr.io/cloud-builders/go` instead? That would give you the `prepare_workspace.inc` and `go_workspace.go` without having to copy them here yourself.

christian-roggia:
I applied all the requested changes, I will test on our Google Cloud environment and then push the changes to this repository.

christian-roggia:
This should be ready for review.

edevil:
What would be the best way to use this until the PR is merged?

jdaily:
@edevil I just cloned the PR's repo:branch  and ran the `gcloud container builds submit . --config=cloudbuild.yaml` command from within dep to get the builder in my environment.

kodova:
Doesn't work with `PROJECT_ROOT` provided by `cloud-builders/go:debian` I tried unsetting the `GOPATH` in env but then the container is no longer able to find dep as its set in `PATH` relative to `GOPATH`.  I was able to fix by using the following snippet.

```
ENV GOPATH=

# Install go dep
RUN /builder/bin/go.bash get -u github.com/golang/dep/cmd/dep
ENV PATH=$GOPATH/bin:$PATH
```

This allows the go environment that dep is running in to be configured by `cloud-builders/go:debian`

red010b37:
The example given in the readme results in
```
Running: dep ensure -v
```

zenthangplus:
Same issue with @red010b37

PR #62, opened 2018-05-22
Add Terraform builder

Thomas-Roper:
Terraform cloud builder with examples of local and GCS backends.

PR #59, opened 2018-05-10
first version of terraform builder

Thomas-Roper:
Working on a builder w/ Terraform for our use, and wanted to share it. 

I looked for an existing builder and found an [issue requesting it](https://github.com/GoogleCloudPlatform/cloud-builders/issues/233) a little while ago. 

Docker image for the builder is from the gsutil builder, rather than the hashicorp official Terraform image. The latest official image is a couple versions behind; and I wanted the gcloud client.

PR #72, opened 2018-06-19
update root cloudbuild to skip sub-builders with no changes from mast…

Thomas-Roper:
A couple of suggestions & a question for the root cloudbuild.yaml for the community project.

1. Most development / pull requests to this project are isolated in impact to just one sub-builder. Triggering builds for all sub-builders is not necessary in the majority of cases. I've implemented a check to compare each sub-builder with GoogleCloudPlatform/cloud-builders-community:master and only trigger sub-builds that are new or have changes. If desired, there's a substitution that can be used to force a trigger for every sub-build.
1. While looking into the check failure for my pull request, I modified the script to write all sub-build output. The current script redirects  output to /dev/null. Normally you could inspect error output by looking at the [builds console](https://console.cloud.google.com/gcr/builds), but if there is an error that prevents the sub-build from being submitted then the output is lost. I've changed the script to direct all output to a file per sub-build, then write the sub-build output only if the sub-build fails. 
1. In my project it looks like the failure is due to all the concurrent builds exhausting a resource-per-user limit. This error prevents the sub-build from being scheduled. Limiting the triggered sub-builds to just those modified in the current branch has a side effect of not hitting this limit in the majority of cases. 
    ```bash
    ERROR: (gcloud.container.builds.submit) RESOURCE_EXHAUSTED: Insufficient tokens for quota 
    'cloudbuild.googleapis.com/get_requests' and limit 'GetRequestsPerMinutePerUser' of service 
    'cloudbuild.googleapis.com' for consumer 'project_number:XXXXXXXX'.
    ``` 
1. The android builder requires a dev key for the build to work. For our work we're not using this builder. With this change: as long as a branch doesn't change the android builder, the root cloudbuilder won't trigger the android build and should pass (assuming the branch changes are legit).

There is a gap introduced by this change. Changes to only the root folder will result in no sub-builds being triggered at all, making the root cloudbuild a no-op. However, this seems like it would be an issue only in a minority of cases. If it does become an issue we can look at triggering all sub-builds if there are any changes in the project root directory. 

As is, someone making a change to the root can use the '_BUILD_ALL' substitution to trigger all the builds; but this will re-introduce the risk of resource exhaustion. It's also possible we could look at using a semaphore to limit the number of concurrent builds; but this seems unnecessary after this change. 

This approach feels like a mild compromise of a continuous integration philosophy; but each sub-folder is virtually a standalone project. In the context of the above issues (managing resource exhaustion, and specialized configuration for some sub-projects) I think this is a reasonable approach.

Finally, a question re: the android builder:
I'm not sure how the root builder can successfully execute this sub-build. Even if an SDK license and build cache were configured as described in the readme, the root script always submits android/cloudbuild.yaml and does not pass through the required environment variables. Looking at the build history, it looks like the Android builder was added March 15, and the Google build-bot started checking before/around April 27. Since then, every pull request has failed the build-bot. 

Is the build-bot running the root cloudbuild? Is there some configuration of the build-bot / root build trigger that addresses the above? Or am I missing something? :)


Thomas-Roper:
@nof20 - I'm relatively new to the community; is there a protocol for requesting a review of a contribution outside of completing the CLA & submitting a pull request? 

Please let me know if there's someone else I should reach out to; I started with you as the top contributor.

Thanks,
-Tom

Thomas-Roper:
Hello @bendory. I'm hoping you might be able to help; is there a protocol for requesting a review of a contribution outside of completing the CLA & submitting a pull request?

This PR addresses 2 issues (#73 , #74 ) that I believe are contributing to the build-bot failures for all the PRs to cloud-builders-community. I've also added a PR for a Terraform builder (#59, #62 ). I've been using it successfully, and it was [requested](https://github.com/GoogleCloudPlatform/cloud-builders/issues/233) in the /cloud-builders project. 

I think both could be useful for others, and I'd value feedback that these additions are consistent with the direction of the cloud builder community. 

Thanks,
-Tom

nof20:
Hey @Thomas-Roper , thanks for your thoughts on this & sorry for my slow response.

I like your suggestion of only triggering builds only for builders which have changed.  @Philmod implemented our CI script - adding him in here.

Thomas-Roper:
Merged changes from master.
I reworked the script into 2 steps; the logic for picking which builders to build has been separated from the step that triggers builds. The intent was to make it easier to understand, and allow tweaks to the build plan logic with minimal impact to the sub-builder handling. 

The build step for triggering builds is back to being gcloud instead of git, which may be marginally better for ongoing maintenance.

Philmod:
Hey @Thomas-Roper,

With an upcoming new feature that will be announced soon, we will have a better way to deal with a repo that contains many builds. I'll keep you updated!

Cheers,
Philmod

PR #75, opened 2018-06-28
Fixed PR builds.

bendory:
Skip android builder which only builds with a local android SDK
installation.
Added .gcloudignore.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -10,13 +10,21 @@ steps:
     failure_file=failure.log
     touch ${failure_file}
 
+
     # Loop through the builders, and build independently.
     for d in */; do
+      # The android builder assumes a local installation of Android SDK which is
+      # not a valid assumption.
+      if [[ "${d}" == "android/" ]]; then
+        echo "Skipping $d."
+        continue
+      fi
+
       echo "Building $d ... "
       (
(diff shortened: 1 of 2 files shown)


cloudbuild.yaml
```diff
@@ -33,4 +55,4 @@ steps:
 
     echo "All builds succeeded."
 
-tags: ['$_PR_NUMBER', '$_REPO_OWNER', '$_REPO', '$_SHA']
+#tags: ['$_PR_NUMBER', '$_REPO_OWNER', '$_REPO', '$_SHA']
```
Philmod: Uncomment.

This is used by a GCF function, we will be able to remove it when we have a better story for PRs.

cloudbuild.yaml
```diff
@@ -25,6 +46,7 @@ steps:
 
     # Check if there is any failure.
     if [[ -s ${failure_file} ]]; then
+      echo
```
Philmod: If you need a line break, can't you add "\n" to the next echo?

bendory: I don't think "\n" works in echo:

```
$ echo "\nfoo\nbar"
\nfoo\nbar
```

Perhaps you mean something else?

PR #79, opened 2018-07-17
Remove FIREBASE_TOKEN env from cloudbuild.yaml

christophersanson:
Building the image throws error: 
secretEnv "FIREBASE_TOKEN" is used without being defined'

Removing FIREBASE_TOKEN env reference fixes


akohen:
#80 will only fix the permission problem when running the container, we still need this to avoid the secretEnv "FIREBASE_TOKEN" is used without being defined' error

PR #81, opened 2018-07-18
Added swift 4.2 cloud builder

Andrewangeta:


ImJasonH:
I think we want the toplevel directory to be just `swift`, and maybe build and tag images with `gcr.io/$PROJECT_ID/swift:4.2` and other versions going forward. That way we can probably share a lot of build logic between `swift:4.2` and `swift:4.3`, etc.

Andrewangeta:
@ImJasonH Gotcha I'll update.

Andrewangeta:
@ImJasonH Any insight as to why it failed?

swift/README.md
```diff
@@ -0,0 +1,10 @@
+# Swift
+
+This build step invokes `swift` commands in [Google Cloud Container Builder](cloud.google.com/container-builder/).
```
ImJasonH: The link needs an `https://` or else it'll link to somewhere on github.com

swift/cloudbuild.yaml
```diff
@@ -0,0 +1,10 @@
+# In this directory, run the following command to build this builder.
+# $ gcloud container builds submit . --config=cloudbuild.yaml
+
+steps:
+- name: 'gcr.io/cloud-builders/docker'
```
ImJasonH: Does it make sense to also publish `gcr.io/$PROJECT_ID/swift:latest`, pointing to `:4.2`? Then when `4.3` or whatever is added, we can point `:latest` at that latest version.

swift/Dockerfile
```diff
@@ -0,0 +1,59 @@
+FROM ubuntu:16.04
+
+# Install related packages and set LLVM 3.8 as the compiler
```
ImJasonH: Can you add a link to the install instructions, in case these fall out of date and someone needs to update them?

swift/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+RUN apt-get -q update && \
+apt-get -q install -y \
```
ImJasonH: Can you format these `RUN` steps a bit to help readability? Indent after newlines, indent the `for` loop below, things like that.

swift/README.md
Andrewangeta: Gotcha wil add this.

swift/cloudbuild.yaml
Andrewangeta: @ImJasonH 
So it depends. Does latest mean the latest Dev snapshot? or does latest mean the latest official release?

swift/Dockerfile
Andrewangeta: 👍

swift/Dockerfile
Andrewangeta: 👍

nof20:
@Andrewangeta @ImJasonH can I do anything to help move this along?

nof20:
I took another quick look at this today - I'm afraid the build now fails because the GPG keys can't be retrieved.  I wonder if it just needs updating?

Andrewangeta:
@nof20 Im going to actually update this now that I found some time. Also the swift.org team has official docker images for swift rather than the whole Dockerfile I'm using now. 
https://github.com/apple/swift-docker

nof20:
Oh awesome, thanks.  At least one of the commits on this PR doesn't have an email/GitHub account with a signed CLA, so if you know you've signed it, might be quicker & easier to close this & start a new one based on the new official Docker image.

Andrewangeta:
@nof20 Yea ill probably do that.

Andrewangeta:
@nof20 Any suggestions of having versions? Should I have multiple `cloudbuild.yaml` and `Docker` files with different swift versions? Maybe have them nested in sub folders? For example if someone wants to use swift 4.2 vs 4.2.1 vs 4.2.2

// README.md
// 4.2/couldbuild.yaml
// 4.2/Dockerfile
// 4.2/examples/

// 4.2.1/couldbuild.yaml
// 4.2.1/Dockerfile
// 4.2.1/examples/

// 4.2.2/couldbuild.yaml
// 4.2.2/Dockerfile
// 4.2.2/examples/

// latest/couldbuild.yaml
// latest/Dockerfile
// latest/examples/

nof20:
I think the approach which minimizes the most duplication is the one the [Angular builder](https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/ng) uses - have a single cloudbuild.yaml but multiple Dockerfiles.  

NB: you can supply multiple tags (-t tag1 -t tag2) in the same Docker build, so you don't need to build the same Dockerfile multiple times, as that one does.

Andrewangeta:
Ahhh I see! looks good. I'll have a PR up shortly.

PR #82, opened 2018-07-24
 Android: use licenses bucket as alternative to single file and creat…

johnlabarge:
Additional features to android builder to make it a little more useful.


johnlabarge:
Not sure what the issue is with the build bot

PR #89, opened 2018-07-29
Removed secretEnv option in firebase

### LINKED ISSUE #88: firebase - build fails
When submitting builds, following error is displayed.
```
ERROR: (gcloud.builds.submit) INVALID_ARGUMENT: invalid build: invalid .secrets field: secretEnv "FIREBASE_TOKEN" is used without being defined
```

`secretEnv: ['FIREBASE_TOKEN']` is not required in cloudbuild.yaml
I will make PR.

iwanariy:
If secretEnv exists, build will fail.
Should fix #88


petomalina:
My bad. Thank you for the fix @iwanariy

PR #93, opened 2018-08-03
Add a crane builder

pdeslaur:
Sample usage of `crane` in a build: https://github.com/pdeslaur/skaffold-test/blob/skaffoldless-test/cloudbuild.yaml. This is useful for deploying to Kubernetes by digest.

### DIFF
--- a/crane/Dockerfile
+++ b/crane/Dockerfile
@@ -1,0 +1,14 @@
+FROM gcr.io/cloud-builders/go as builder
+
+# Only allow static linking, so that crane can be used in a different
+# environment.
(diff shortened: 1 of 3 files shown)


ImJasonH:
There's actually already an image built at `gcr.io/go-containerregistry/crane` though this isn't really advertised anywhere, or rebuilt on changes.

pdeslaur:
Oh very cool! Should we just close this PR then or do you think there's still some value in having it here?

ImJasonH:
Nah I think we can just close it.

PR #96, opened 2018-08-15
Add Kustomize Custom Builder

hayesgm:
This patch adds a Kustomize cloud builder. The builder (largely based off of helm) easily allows users of [kustomize](https://github.com/kubernetes-sigs/kustomize) to build and apply kustomize templates in Cloud Build. We've opted to add an entrypoint script that allows the user to apply the kustomize script when he/she sets the env var `APPLY=true`. We use the same variables for getting gcloud settings as Helm and gcloud. Additionally, we've added instructions for using non-standard kustomize runs.

### DIFF
--- a/kustomize/cloudbuild.yaml
+++ b/kustomize/cloudbuild.yaml
@@ -1,0 +1,9 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/$PROJECT_ID/kustomize', '.']
+- # Verify kustomize has been built correctly
+  name: 'gcr.io/$PROJECT_ID/kustomize'
+  entrypoint: 'kustomize'
+  args: ['version']
+
+images: ['gcr.io/$PROJECT_ID/kustomize']
--- a/kustomize/Dockerfile
+++ b/kustomize/Dockerfile
@@ -1,0 +1,21 @@
+FROM gcr.io/cloud-builders/gcloud
+
+ENV VERSION v1.0.6
+
+COPY kustomize.bash /builder/kustomize.bash
+
+RUN apt-get update && \
+    apt-get install -y wget && \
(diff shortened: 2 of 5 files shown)


sebgoa:
@hayesgm you need to sign the CLA for this to go forward.

I created #102 but I am not attached to it. I do like the way I used the `wget` builder in the `cloudbuild.yaml` file to pull the kustomize binary that makes the `Dockerfile` really simple.

hayesgm:
Yeah, I have signed the CLA and got the confirmation email. Not sure why the Google bot has not updated as such.

The other thing I like about this PR is that it allows you to do `kubectl apply` without having to change the entry point. I think that could be powerful.

sebgoa:
@hayesgm just force push to your branch to trigger the build again

hayesgm:
I signed it!

hayesgm:
Let's see if that kicks off the verification. In either case, happy to go with either strategy. But again, I do think a nice feature in either will be to allow `kubectl apply` without changing the entrypoint.

sebgoa:
@ImJasonH Hi, any chance to get a nudge here to see what's up with the CLA ? thanks

hayesgm:
I signed it! (for a third time, with a different email address-- let's see if this works)

kustomize/cloudbuild.yaml
```diff
@@ -0,0 +1,5 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/$PROJECT_ID/kustomize', '.']
+
```
ImJasonH: Can you add a step that runs the builder, even if it's just `kustomize version`, to make sure the tool is installed correctly?

hayesgm:
Rebased off of master and added a step to verify that the image correctly built kustomize.

PR #99, opened 2018-08-16
add traceroute tool

leeonlee:


traceroute/cloudbuild.yaml
```diff
@@ -0,0 +1,11 @@
+steps:
+# Build the image to your GCR using the included Dockerfile
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '-t', 'gcr.io/$PROJECT_ID/traceroute', '.']
+
+# Print version information
+- name: 'gcr.io/$PROJECT_ID/traceroute'
+  args: ['-V']
```
ImJasonH: I think you could just move the example `traceroute google.com` here too, it's simple enough it doesn't really need to be in the `examples/` directory.

nof20: @leeonlee What do you think?

leeonlee: Done.

PR #103, opened 2018-08-20
Add kubeless builder

sebgoa:
I am not entirely happy about the size of this image. it is huge because of the underlying `kubectl` image.

so maybe let's treat this as WIP


sebgoa:
Actually, I should have closed this PR as we are migrating to knative. We don't need kubeless anymore.

PR #110, opened 2018-09-18
use build args to source helm version

curtisallen:
Allows helm to be updated via docker build-arg, this will allow updates of the helm version without modifying the dockerfile. 

👀 @rimusz

PR #111, opened 2018-09-19
Improve Tillerless Helm and add RBAC setup to readme.

rimusz:
Improve Tillerless Helm use case and add RBAC setup example to helm readme.
Incorporate #110, which upgrades Helm to v2.10.0 and allows helm to be updated via docker build-arg, this will allow updates of the helm version without modifying the dockerfile.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -37,21 +37,54 @@ Then, `kubectl` and consequently `helm` will have the configuration needed to ta
 
 To build this builder, run the following command in this directory.
 
-    $ gcloud builds submit . --config=cloudbuild.yaml
+    gcloud builds submit . --config=cloudbuild.yaml
+
+You can also build this builder setting `Helm` version via in `cloudbuild.yaml`, no need to do that in `Dockerfile` anymore.
+
+    args: ['build', '--tag=gcr.io/$PROJECT_ID/helm', '--build-arg', 'HELM_VERSION=v2.10.0', '.']
 
 ## Using Helm
 
 This builder supports two install options of Helm:
-* The default one when the `tiller` gets installed into your GKE cluster (oh all those `tiller` security issues)
-* `Tillerless Helm`, I wrote a [blog post](https://rimusz.net/tillerless-helm/) of using Helm local [tiller plugin](https://github.com/rimusz/helm-tiller) which solves all those `tiller` security issues, as `tiller` runs outside the GKE cluster.
+* The default one when the `tiller` gets installed into your GKE cluster.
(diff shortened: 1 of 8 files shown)


helm/README.md
```diff
@@ -37,21 +37,54 @@ Then, `kubectl` and consequently `helm` will have the configuration needed to ta
 
 To build this builder, run the following command in this directory.
 
-    $ gcloud builds submit . --config=cloudbuild.yaml
+    gcloud builds submit . --config=cloudbuild.yaml
+
+You can also build this builder setting `Helm` version via in `cloudbuild.yaml`, no need to do that in `Dockerfile` anymore.
+
+    args: ['build', '--tag=gcr.io/$PROJECT_ID/helm', '--build-arg', 'HELM_VERSION=v2.10.0', '.']
```
curtisallen: thanks for updating the readme

helm/helm.bash
```diff
@@ -44,15 +44,20 @@ fi
...
 
-# check if Tillerless value 'TILLERLESS=true' is provided then install the plugin
-# and run "helm tiller run $@"
+# check if Tillerless value 'TILLERLESS=true' is provided then install and start the plugin
 if [ "$TILLERLESS" = true ]; then
   echo "Installing Tillerless plugin"
   helm plugin install https://github.com/rimusz/helm-tiller
+  echo "Starting Tillerless plugin"
+  helm tiller start-ci "$TILLER_NAMESPACE"
+  echo
```
curtisallen: did you mean to leave in this blank echo?

helm/helm.bash
```diff
@@ -44,15 +44,20 @@ fi
...
+  export HELM_HOST=localhost:44134
   if [ "$DEBUG" = true ]; then
-      echo "Running: helm tiller run $@"
+      echo "Running: helm $@"
   fi
-  helm tiller run "$@"
+  helm "$@"
+  sleep 2
```
curtisallen: why sleep? seems like this could be buggy for some users depending on their GKE config

helm/README.md
rimusz: of course

helm/helm.bash
rimusz: yes, to make logs more readable

helm/helm.bash
rimusz: I put it just in case, but I did not find any issues without the `sleep 2`, so sure going to remove it

rimusz: done

rimusz:
ping :)
can we please get this one merged?

PR #115, opened 2018-10-09
Update hugo to 0.49

takuyaohashi:
- Update hugo to 0.49


takuyaohashi:
@googlebot 

I signed it!

PR #119, opened 2018-10-20
[Android Builder] Check executable before script

dlackty:
Hi, first of all, thanks for making Android builder for Cloud Build.

While my team were testing this builder, we found that the list package script just failed silently, and thus want to add a warning in this PR.


nof20:
Looks good.  Thanks!

PR #123, opened 2018-10-31
Updated Dockerfile and firebase.bash

classhopper:
After the latest release of Firebase tools (`6.0.0`) the installation command `npm i firebase-tools` in the Dockerfile did not work in conjunction with the `export PATH=$PATH:/node_modules/firebase-tools/bin` in the `firebase.bash`. Changing the installation command to `npm install -g firebase-tools` and removing the `export PATH` part in `firebase.bash` seems it does the job.

### DIFF
--- a/firebase/Dockerfile
+++ b/firebase/Dockerfile
@@ -1,6 +1,7 @@
 FROM gcr.io/cloud-builders/npm
 
-RUN npm i firebase-tools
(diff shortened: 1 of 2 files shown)


nof20:
Hey @classhopper , thanks a lot for the contribution.  Looks like we just merged an equivalent change (#128) which does the same thing, so I think we can close this.

PR #124, opened 2018-11-05
Add npm puppeteer container

sankaritan:
Adding a container for [puppeteer](https://github.com/GoogleChrome/puppeteer/) and npm.

This is based on the existing container `yarn-puppeteer` but using npm instead.

I also bumped puppeteer version to the latest one (1.9.0).

### DIFF
--- a/npm-puppeteer/Dockerfile
+++ b/npm-puppeteer/Dockerfile
@@ -1,0 +1,24 @@
+ARG NODE_VERSION
+FROM node:$NODE_VERSION-alpine
+
+# Installs latest Chromium (64) package.
+RUN apk update && apk upgrade && \
+    echo @edge http://nl.alpinelinux.org/alpine/edge/community >> /etc/apk/repositories && \
+    echo @edge http://nl.alpinelinux.org/alpine/edge/main >> /etc/apk/repositories && \
+    apk add --no-cache \
+      chromium@edge \
+      nss@edge \
+      git
+
+# Tell Puppeteer to skip installing Chrome. We'll be using the installed package.
+ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true
+ENV CI true
+
+# Puppeteer v1.4.0 works with Chromium 68.
+RUN npm install puppeteer@1.9.0
+
+# It's a good idea to use dumb-init to help prevent zombie chrome processes.
+ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.0/dumb-init_1.2.0_amd64 /usr/local/bin/dumb-init
+RUN chmod +x /usr/local/bin/dumb-init
+
+ENTRYPOINT ["dumb-init", "--", "npm"]
--- a/npm-puppeteer/cloudbuild.yaml
+++ b/npm-puppeteer/cloudbuild.yaml
@@ -1,0 +1,67 @@
+# In this directory, run the following command to build this builder.
+# $ gcloud builds submit . --config=cloudbuild.yaml
+#
+
+steps:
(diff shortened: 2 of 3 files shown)


sankaritan:
I signed it!

palmerj3:
👍

npm-puppeteer/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+      chromium@edge \
+      nss@edge \
+      git
+
+# Tell Puppeteer to skip installing Chrome. We'll be using the installed package.
+ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true
+ENV CI true
+
+# Puppeteer v1.4.0 works with Chromium 68.
+RUN npm install puppeteer@1.9.0
```
javoire: ops, small discrepancy between code and comment re version! Also are you sure 1.9.0 works?

npm-puppeteer/Dockerfile
```diff
@@ -0,0 +1,24 @@
+ARG NODE_VERSION
+FROM node:$NODE_VERSION-alpine
+
+# Installs latest Chromium (64) package.
```
javoire: change `(64)` to `(68)`!

npm-puppeteer/Dockerfile
javoire: also I had to add extra packages due to a "missing symbols" error when running with chromium 68. See this https://github.com/GoogleCloudPlatform/cloud-builders-community/pull/125#issue-228746411

javoire: after further investigation, the "missing symbols" error appears in node 6,8 and 9 but not in 10, for puppeteer 1.9.0.

nof20:
Hey @sankaritan @javoire - are you happy with the current state of this PR?  Or do you need to iterate a bit further?

npm-puppeteer/cloudbuild.yaml
```diff
@@ -0,0 +1,67 @@
...
+# hello_world
+- name: 'gcr.io/$PROJECT_ID/npm-puppeteer:latest'
+  args: ['install']
+  dir: 'examples/hello_world'
+- name: 'gcr.io/$PROJECT_ID/npm-puppeteer:latest'
+  args: ['test']
+  dir: 'examples/hello_world'
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '.']
+  dir: 'examples/hello_world'
```
javoire: We should replace this with similar to this: https://github.com/javoire/cloud-builders-community/blob/a286e45988888e17be7e39f6ffb0dc1f3404db27/yarn-puppeteer/cloudbuild.yaml#L47-L74

e.g. test all versions against the example app

javoire:
Hey, apologies for the late reply!

We should fix the comments and add the extra tests I just commented, to make sure this works for all those node versions.

Cheers.

nof20:
@sankaritan @javoire - @tomchinery was kind enough to submit another npm-puppeteer builder, which suggests to me that people really do need this.

I'm tempted to merge this as it is so people can iterate on it.  Is that OK?

palmerj3:
👍

codrienne:
I'm going through the backlog of open PRs. 

As @nof20 pointed out, @tomchinery already has https://github.com/GoogleCloudPlatform/cloud-builders-community/pull/236/files open. I'm missing context, but I checked out the PR and  cloudbuild.yaml no longer builds. :( I'm inclined to close this PR and let the new PR with tomchinery@ go under review. Please re-open if you think otherwise.

PR #125, opened 2018-11-06
Use puppeteer 1.4.0

javoire:
and add chromium 68 missing symbols fix: https://github.com/GoogleChrome/puppeteer/issues/3019#issuecomment-417227105


javoire:
I signed it!

PR #127, opened 2018-11-06
Updates Android builder to be more encapsulated, configurable and complete

rharter:
This is a fairly substantial change to the Android builder, so I'd welcome discussion.

The changes to the Android `Dockerfile` and sample `cloudbuild.yaml` included here are based on [my prototype](https://github.com/pixiteapps/android-cloud-build).

There are a couple main points that I think this PR addresses that will make integrating Google Cloud Build into an Android workflow much easier for new users.

## 1. Android images should be flexible.

The original image configuration was cumbersome and highly customized, instead of meeting the needs of most users and allowing them to optimize from there.

With this change user's can easily (with 1 argument) configure the generated Android image based on their target API.  This is the most common thing that Android developers need to change, since other missing packages will be downloaded on demand.

From here it's easy to optimize the build by adding dependent packages, but user's don't need to do that from the start.

For users needing the NDK for their builds, the `cloudbuild-ndk.yaml` takes care of using the appropriate Dockerfile to include that.

## 2. Build settings live in the build config

Originally the build cache caching was in two places, the docker image (via the `gradle-build` command), and the `cloudbuild.yaml` configuration file (via the `gsutil` download at the right path).

This is cumbersome, as changes to the caching require both a new image and updates to the build script, but also require the build script to use commands that are different from the commands users run on their local machines.

It also meant that the build config would be inflated and compressed for every build step that ran gradle.

By removing the `gradle-build` script, and truly making caching part of the build configuration via the `cloudbuild.yaml` file, all of these problems are solved, allowing the images to be used almost without modification.


android/examples/build-test-deploy.yaml
```diff
@@ -0,0 +1,232 @@
+steps:
+
+#
+# Get a build number
```
rharter: While this section isn't imperative for Android builds, incrementing build numbers is something that other major CI solutions offer and are used for many Android builds (for things like APK version codes).  This allows this sample to be used with extremely little modification by developers to try out Cloud Build and get started quickly.

android/examples/build-test-deploy.yaml
```diff
@@ -0,0 +1,232 @@
...
+- name: 'gcr.io/cloud-builders/gsutil'
+  id: save_env
+  args: ['cp', '/config/buildnum', 'gs://${_CONFIG_BUCKET}/buildnum']
+  waitFor: ['setup_env']
+  volumes:
+  - name: 'config'
+    path: '/config'
+
+#
+# Extract the cache
```
rharter: Instead of having cache configuration live both in the `cloudbuild.yaml` **and** the `gradle-build` file, this section moves it to the config file, making the docker images far more flexible.

android/examples/build-test-deploy.yaml
```diff
@@ -0,0 +1,232 @@
...
+  args:
+  - '-c'
+  - |
+    tar xpzf /build_cache/cache.tgz -C /build_cache || echo "No cache found."
+  volumes:
+  - name: 'build_cache'
+    path: '/build_cache'
+
+#
+# Decrypt secret files
```
rharter: Another thing that's not imperative for Android but is an extremely common use case as Android builds require things like keystores in order to deploy to Play.

android/examples/build-test-deploy.yaml
```diff
@@ -0,0 +1,232 @@
...
+# Deploy to Google Play if this is the master branch.
+#
+- name: 'gcr.io/$PROJECT_ID/android:28'
+  id: deploy_to_play
+  entrypoint: 'bash'
+  args:
+  - '-c'
+  - |
+    source build_environment.sh
+    [[ "$BRANCH_NAME" == "master" ]] && ./gradlew --stacktrace -Dtrace :app:publishReleaseApk || echo "Skipping Play Store distribution..."
```
rharter: Would love to see either configurable workflows ala CircleCI, or at least simple branch filtering to avoid this type of thing.

tar/README.md
```diff
@@ -0,0 +1,28 @@
+# tar
```
rharter: This mirrors the "command images" found in the official `cloud-builders` repo.

android/cloudbuild.yaml
```diff
@@ -1,24 +1,27 @@
 steps:
-- name: 'gcr.io/cloud-builders/gsutil'
-  entrypoint: "bash"
-  args: 
-  - '-c'
-  - |
-    gsutil rsync -d -r "${_LICENSES_BUCKET}" licenses || exit 0
-  id: copy-licenses
+
```
nof20: Nit: remove trailing whitespace.
```suggestion

```

android/cloudbuild.yaml
```diff
@@ -1,24 +1,27 @@
...
+  - '--file=Dockerfile-sdk'
   - '.'
   - '--cache-from'
-  - 'gcr.io/$PROJECT_ID/android-sdk-base'
-  waitFor: 
-  - copy-licenses
-images: ['gcr.io/$PROJECT_ID/android-builder']
+  - 'gcr.io/$PROJECT_ID/android:base'
+
+images:
```
nof20: Nit: remove trailing whitespace.
```suggestion
images:
```

android/README.md
```diff
@@ -1,109 +1,116 @@
...
+gcloud builds submit --config=cloudbuild-ndk.yaml --substitutions=_ANDROID_VERSION=28
+# ...
+# Step #2: Successfully tagged gcr.io/project-id/android:28-ndk-r17b
+```
+
+### 2. Create a cloudbuild.yaml file
+
+Create a `cloudbuild.yaml` file based on your needs and save it in the root of your project.
+
+Check out the [complete sample](samples/build-test-deploy.yaml) for some ideas of how to accomplish this.
```
nof20: s/sample/example/
```suggestion
Check out the [complete example](examples/build-test-deploy.yaml) for some ideas of how to accomplish this.
```

tar/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM launcher.gcr.io/google/ubuntu16_04
```
nof20: What about using the Alpine base image instead?  It has minimal packages installed, so loads much faster and has a far smaller attack surface.  It's also cached on Cloud Build workers, and reduces build time from 35 seconds to 14 seconds.
```suggestion
FROM alpine
```

tar/Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+
+RUN apt-get update && \
```
nof20:
```suggestion
RUN apk update && \
```

tar/Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+    apt-get -y install tar bzip2 gzip
```
nof20:
```suggestion
    apk install tar bzip2 gzip
```

android/examples/build-test-deploy.yaml
nof20: Understood.  You can set up [different builds to run from different branches](https://cloud.google.com/cloud-build/docs/running-builds/automate-builds) in Cloud Build, but can't introspect the branch the build is running from.

android/examples/build-test-deploy.yaml
nof20: Kicking the tires & getting started quickly sounds like an important goal, especially since mobile developers might be less familiar with containers in general.

Is it possible to add a simpler example which doesn't require setup of e.g. GCS buckets, Cloud KMS keys, etc.?  In other places we download and build sample code used in a quick start tutorial.

android/examples/build-test-deploy.yaml
nof20: Sounds great.

android/examples/build-test-deploy.yaml
nof20: Agreed

tar/README.md
nof20: Sounds good

android/README.md
```diff
@@ -1,109 +1,116 @@
...
+| `_CONFIG_BUCKET` | my-build-config |
+
+
+# Contact Us
 
 Please use [issue tracker](https://github.com/GoogleCloudPlatform/android-cloud-build/issues)
 on GitHub to report any bugs, comments or questions regarding SDK development.
 
 We welcome all usage-related questions on [Stack Overflow](http://stackoverflow.com/questions/tagged/google-android-cloud-build)
 tagged with `google-cloud-container-builder`.
```
nof20:
```suggestion
tagged with `google-cloud-build`.
```

android/Dockerfile-sdk
```diff
@@ -0,0 +1,8 @@
+# we use a gcr.io image and not openjdk:8-jdk-slim because it loads faster in the google Google Cloud Build environment
```
nof20: Nit: suggest you remove this line and keep only the copy in `Dockerfile-base`, where it's in the right context.  Anyone building using this Dockerfile would have to pull `gcr.io/$project_id/android:base` from Container Registry.

android/examples/build-test-deploy.yaml
rharter: Yeah, I like that idea. I'll work on getting a simpler sample together.

rharter:
I've updated this as requested.  Please let me know if there's anything else needed, @nof20.

android/examples/build-test-deploy.yaml
johnlabarge: This is very cool.

johnlabarge:
@rharter  Thanks so much for the contributions!

rharter:
@johnlabarge absolutely!  

I have run into issues with the alpine image as I've exercised this some more (specifically for Android), so I'd hold off a little while I work that out.

I'm happy to close this and reopen when that's solved, or I can leave this "on hold".

nof20:
@rharter Thanks for all these changes.  If you're having compatibility problems with Alpine in the Android images, fall back to gcr.io/cloud-builders/javac.  (My suggestion to use Alpine was just for the tar image because it's so simple - sorry if that was misleading.)

CLA is now fully approved too.  I'm good to approve this as it stands.

rharter:
Thanks, @nof20. I've reverted the Android image to gcr.io/cloud-builders/javac and added some scripts to the image to help with Android builds.  I'm currently using this image and script in my project, so can verify that it is working well.

nof20:
Thanks!

PR #135, opened 2018-11-20
Remove cron-helper

nof20:
[Cloud Scheduler](https://cloud.google.com/scheduler/) is now launched in Beta, and it's a way better solution than this Rube Goldberg-style App Engine solution...


nof20:
Cancelling this after discussion.

PR #138, opened 2018-11-21
 Add nix-build builder

ImJasonH:
This adds a builder to use the [Nix build system](https://nixos.org/nix/manual/#sec-nix-build), which executes builds described in the Nix config language, and an example script to produce a layered Docker image based on [these instructions](https://grahamc.com/blog/nix-and-layered-docker-images).

cc @rafikk

### DIFF
--- a/nix-build/Dockerfile
+++ b/nix-build/Dockerfile
@@ -1,0 +1,6 @@
+FROM nixos/nix
+
+RUN nix-channel --add https://nixos.org/channels/nixpkgs-unstable nixpkgs
+RUN nix-channel --update
+
+ENTRYPOINT ["nix-build"]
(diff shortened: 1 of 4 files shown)


clrprod:
You'll likely want to pass the thing to build as an arg via the --argstr flag to nix-build. In order for default.nix to understand it, you'll need to make it a function

i.e.
make default.nix look like
{ docker_path }:
... rest of your existing file

and build it with nix-build --argstr docker_path gcr.io/...

ImJasonH:
@clrprod That was exactly the hint I needed, thanks, it works now! 🎉

clrprod:
Huzzah!

PR #139, opened 2018-11-21
Add makisu example

ImJasonH:
This demonstrates using Uber's [`makisu`](https://github.com/uber/makisu) in Google Cloud Build to produce Docker images, instead of using `docker build`.

Unfortunately, seemingly due to an issue to consume the GCB builder service account's Application Default Credentials, this still has to depend on the Docker daemon and `--load` the image before pushing it. If we can fix that, we could replace `--load` with `--push=gcr.io` and push the image directly to the registry.

The error when using `--push=gcr.io` is:

```
2018/11/21 19:19:52 Command failure: failed to push image: failed to push image: check manifest exists for image gcr.io/my-project/built-with-makisu:latest: check manifest exists: HEAD https://gcr.io/v2/my-project/built-with-makisu/manifests/latest 401
```

cc @yiranwang52

### DIFF
--- a/makisu/Dockerfile
+++ b/makisu/Dockerfile
@@ -1,0 +1,4 @@
+FROM ubuntu
+
+RUN apt-get update -y
+RUN apt-get install -y ca-certificates #!COMMIT
--- a/makisu/README.md
+++ b/makisu/README.md
@@ -1,0 +1,5 @@
+# Makisu
+
+This directory contains an example of using [Uber's Makisu
+tool](https://github.com/uber/makisu) to produce Docker images in
+Google Cloud Build.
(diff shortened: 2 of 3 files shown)


yiranwang52:
ack.

evelynl94:
@ImJasonH did you configure docker using https://cloud.google.com/container-registry/docs/advanced-authentication?

If so, you might need to add your token in a [registry configuration](https://github.com/uber/makisu#configuring-docker-registry) and pass it in to Makisu. You can create a token here: https://console.cloud.google.com/apis/credentials?project=$PROJECT

I don't know if gcr allows creating public repos that anyone can push to without auth. If that is the case, I can improve Makisu's push default on that.

PR #151, opened 2018-12-13
Helm Tillerless tweaks

afirth:
## Why

* The current implementation of tillerless helm works well, but downloads the tiller binary and tillerless plugin every build. This causes unnecessarily long build times. Most plugin functionality is not needed for non-interactive use.
* The helm default versions are out of date
* The README for RBAC clusters is incomplete and a bit roundabout

## What this does

* Install tiller locally in the Dockerfile (it was already pulled but discarded from the tarball)
  - adds about 23 MB to the compressed image
* Modify helm.bash to use the locally installed tiller in a similar manner to the plugin which was previously used
  - start tiller in the background
  - tell helm to connect to the local tiller
  - pkill tiller
* Bump helm to 2.12.0
* Update the README

## Benefits
* Speeds up build startup on standard-machine from 70sec to 10sec

### DIFF
--- a/helm/Dockerfile
+++ b/helm/Dockerfile
@@ -1,6 +1,6 @@
 FROM gcr.io/cloud-builders/gcloud
 
-ARG HELM_VERSION=v2.10.0
+ARG HELM_VERSION=v2.12.0
 
 COPY helm.bash /builder/helm.bash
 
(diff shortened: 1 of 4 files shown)


nof20:
Looks good, thanks for your contribution.

PR #158, opened 2019-01-06
Firebase builder: Step missing to set $TOKEN used for encrypting

rav84:
The glcoud encrypt command uses $TOKEN variable, which is not set anywhere.


rav84:
I signed it.

nof20:
Good spot.  Thanks!

Issue #157, opened 2019-01-05
firebase - instructions using login:ci and kms are unnecessary


leighmcculloch:
The instructions on the `firebase` image state that the user should use the `firebase login:ci` command locally to generate a token, encrypt the token with KMS, then store that encrypted token in the `cloudbuild.yaml` file. These steps are unnecessary and incur additional costs because of the use of KMS to encrypt the secrets.

You can instead directly grant the Build service account user (it'll have the name `<project-number>@cloudbuild.gserviceaccount.com`) the following roles and it will be able to perform firebase deploys, at least of the 'Firebase Hosting' product. Some additional roles might need adding for other Firebase products.

1. Cloud Build Service Account
2. Firebase Admin
3. API Keys Admin

You also need to enable these APIs for the project if they aren't already enabled:

1. [Cloud Resource Manager API](
https://console.developers.google.com/apis/api/cloudresourcemanager.googleapis.com)
2. [Firebase Management API](
https://console.developers.google.com/apis/api/firebase.googleapis.com)
3. [Firebase Hosting API](
https://console.developers.google.com/apis/api/firebasehosting.googleapis.com)

Is there a reason the `firebase login:ci` approach is promoted here? Is it preferred for some reason, or just outdated?

ImJasonH:
I suspect it's just outdated. That's great news that these steps aren't necessary! Care to phrase this in terms of a PR we would happily merge? 🙏

PR #159, opened 2019-01-08
firebase: allow for IAM or token based auth

MylesBorins:
The current approach for authenticating with FireBase required
generating and manually managing a key via KMS. This update documents
how to authenticate via IAM roles instead, which is much simpler.

`firebase.bash` has been updated to allow for the `$FIREBASE_TOKEN` to
be committed, but will still support it for the legacy use case

Fixes: https://github.com/GoogleCloudPlatform/cloud-builders-community/issues/157

### DIFF
--- a/firebase/README.md
+++ b/firebase/README.md
@@ -8,6 +8,22 @@ command](https://firebase.google.com/docs/cli/#command_reference).
 
 ## Usage
 
+### With IAM roles
+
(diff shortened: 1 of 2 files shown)


nof20:
Looks good.  Thanks!

leighmcculloch:
@MylesBorins Thanks for updating the instructions and the image!

PR #161, opened 2019-01-09
Helm Tillerless exit code

kszalontai:
Tillerless Helm fails silently: exit code is not specified in the script, so it's the exit code of the last command, which is `pkill tiller` (always 0)

This fix solves this issue, also bumps Helm version.


rimusz:
LGTM

nof20:
Thank you!

PR #173, opened 2019-01-26
Adding Singularity (container) Google Builder Recipes

vsoch:
I'm working with @wkharold on some tutorials to use Singularity with Google Cloud Builder, and we thought it would be great to contribute recipes here! (see issue https://github.com/wkharold/community/issues/10https://github.com/wkharold/community/issues/10). specifically, this pull request will add a singularity folder with subfolders to create a builder (create-builder) and then run the builder (run-builder) to build and send the container to Google Cloud storage. Please let us know if you'd like to see any changes, or bring up ideas for discussion. Looking forward to it, and happy weekend! :)


nof20:
Thanks a lot for your submission but, after some consideration, I merged another singularity builder with similar functionality.  It would be great if you could take a look at that solution and suggest improvements.

vsoch:
No worries, it was a suggestion by @wkharold and a (similar) tutorial is published here - https://cloud.google.com/community/tutorials/singularity-containers-with-cloud-build probably best to have variation! I'd b happy to give some quick feedback on the tutorial that you merged.

 - the Dockerfile doesn't use a multistage build. [Here is an example](https://github.com/singularityhub/singularity-docker/blob/3.1.1-slim/Dockerfile) of one that does. The difference in size can go from ~314 to ~40 mb [see builds here](https://hub.docker.com/r/singularityware/singularity/tags).
 - centos (historically) has more random issues than debian, so I wouldn't have used it as a base. 
 - There isn't any logic / organization to where the images end up in storage.

Looking forward to improvements!

PR #174, opened 2019-01-28
Added GCS cache builder

kszalontai:
### Google Cloud Storage cache builder
You can increase the speed of your build with this builder [using a GCS bucket as a cache](https://cloud.google.com/cloud-build/docs/speeding-up-builds#caching_directories_with_google_cloud_storage).

### DIFF
--- a/gcs-cache/README.md
+++ b/gcs-cache/README.md
@@ -1,0 +1,29 @@
+[GCS cache]: https://cloud.google.com/cloud-build/docs/speeding-up-builds#caching_directories_with_google_cloud_storage
+[Parallel Composite Uploads]: https://cloud.google.com/storage/docs/gsutil/commands/cp#parallel-composite-uploads
+
+# Google Cloud Storage cache builder
+To increase the speed of your build, you [can use a GCS bucket as a cache][GCS cache] with this builder.
+
+
+## Building this builder
+To build this builder, run the following command in this directory.
+
+```bash
+gcloud builds submit . --config=cloudbuild.yaml
+```
+
+## Using this builder
+* Saving cache to the bucket with the `-s` flag
+* Restoring cache from a bucket with the `-r` flag
+
+| Option                | Description                                                         |
+| --------------------- | ------------------------------------------------------------------- |
+| `-r`                  | Mode: restore                                                       |
+| `-s`                  | Mode: save                                                          |
+| `-f file1 -f file2`   | File(s) to save, mandatory when saving                              |
+| `-c gradle-cache.zip` | The compressed file name, optional, default `cache.zip`             |
+| `-t 100`              | [Parallel Composite Uploads] threshold size, optional, default `50` |
+
+The `GCS_BUCKET` environment variable must be passed to the build step in the `env` parameter.
+
+Check the [examples](examples) folder for a common usage of this builder.
--- a/gcs-cache/Dockerfile
+++ b/gcs-cache/Dockerfile
@@ -1,0 +1,11 @@
+FROM gcr.io/cloud-builders/gcloud-slim
+
+RUN apt-get -y update && \
+    apt-get -y install zip && \
+    apt-get -y install unzip
+
+COPY gcs-cache.bash /builder/gcs-cache.bash
+
+RUN chmod +x /builder/gcs-cache.bash
+
+ENTRYPOINT ["/builder/gcs-cache.bash"]
--- a/gcs-cache/cloudbuild.yaml
+++ b/gcs-cache/cloudbuild.yaml
@@ -1,0 +1,5 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/$PROJECT_ID/gcs-cache', '.']
+
+images: ['gcr.io/$PROJECT_ID/gcs-cache']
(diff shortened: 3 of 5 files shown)


gcs-cache/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+| Option                | Description                                                         |
+| --------------------- | ------------------------------------------------------------------- |
+| `-r`                  | Mode: restore                                                       |
+| `-s`                  | Mode: save                                                          |
+| `-f file1 -f file2`   | File(s) to save, mandatory when saving                              |
+| `-c gradle-cache.zip` | The compressed file name, optional, default `cache.zip`             |
+| `-t 100`              | [Parallel Composite Uploads] threshold size, optional, default `50` |
+
+The `GCS_BUCKET` environment variable must be passed to the build step in the `env` parameter.
```
rharter: Have you considered making this also available via a command parameter? It might be a little cleaner in the `cloudbuild.yaml` file to be able to have all parameters provided via parameters instead of some provided via params and some via env variables.

```yaml
- name: 'gcr.io/$PROJECT_ID/gcs-cache'
  args: [ '-s',
          '-b', 'gs://$PROJECT_ID-cache'
          '-f', '.gradle',
          '-c', 'gradle-cache.zip']
```

rharter:
Didn't want to duplicate, but I had been working on the same problem, so shared in #188.  Happy to discuss pros and cons and if these might be unified (or one better than the other).

oliverxchen:
Thanks for both of these! A good way to handle caching will make Cloud Build a lot better.

I'll kick off a discussion on pros/cons with what I like about these two approaches:

#174 (this PR):

* Uses parallel composite uploads
* Uses a single docker image 

#188 (@rharter's PR):

* ability to checksum. That's pretty critical for some types of builds

rharter:
Thanks, @oliverxchen!  Those are some good points.

Here are a few of my other thoughts:

### Separate Docker Image

I actually like having this in two separate steps, since each has separate parameters.  For instance, the save operation has things like files to include, and the composite upload threshold, which don't apply to restore.  This, I think, can make things a little confusing.

Separate steps also acts a bit more like other tools (that I'm used to, at least), like CircleCI and Bitrise. I, personally, find this to be easier to read and reason about in the build script, but understand how that is purely personal preference.

As the difference in the docker containers is just the scripts, I don't expect it to add much in download time to the builders.

### `=` based arguments

I know this is not necessarily standard in bash scripts, but I felt that, short of being able to pass args as key/value pairs in the yaml, being able to have arguments keys and values in a single item was important because of how they are set.

As @kszalontai shows here, you can make this readable by formatting your build script, but that locks you into formatting, which is difficult in other circumstances.

```yaml
- name: 'gcr.io/$PROJECT_ID/gcs-cache'
  args: [ '-s',
          '-b', 'gs://$PROJECT_ID-cache'
          '-f', '.gradle',
          '-c', 'gradle-cache.zip']
```

This doesn't really work, for instance, using Yaml's list support:

```yaml
- name: 'gcr.io/$PROJECT_ID/gcs-cache'
  args: 
  - '-s'
  - '-b'
  - 'gs://$PROJECT_ID-cache'
  - '-f'
  - '.gradle'
  - '-c'
  - 'gradle-cache.zip'
```

That's why I chose go to with `=` based key/value pairing, so that arguments could be specified and easily read on one line:

```yaml
- name: gcr.io/$PROJECT_ID/save_cache
  args:
  - --bucket=gs://$PROJECT_ID-cache
  - --key=build-cache-$( checksum build.gradle )
  - --path=.gradle/caches
  - --path=.gradle/wrapper
```

What I would love is to be able to pass these as a Yaml map, but that's not currently possible, AFAIK.

```yaml
- name: gcr.io/$PROJECT_ID/save_cache
  args:
    bucket: gs://$PROJECT_ID-cache
    key: build-cache-$( checksum build.gradle )
    paths:
    - .gradle/caches
    - .gradle/wrapper
```

### All args from same source

This PR requires a `GCS_BUCKET` env variable, which means some arguments are coming from the environment and some passed directly via the `args` map.

As I commented before, this should be unified, or optional.  I like the idea of allowing for this via environment variables, as you can using Yaml's snippet support to share the env between all of your steps, but if you don't use that it just makes the build scripts messier.

```yaml
- name: gcr.io/$PROJECT_ID/save_cache
  args:
  - --bucket=gs://$PROJECT_ID-cache
  - --key=build-cache-$( checksum build.gradle )
  - --path=.gradle/caches
  - --path=.gradle/wrapper
  <<: &env
    env:
    - 'TERM=dumb'
    - 'JAVA_TOOL_OPTIONS="-Xmx4g"'
    - 'GRADLE_USER_HOME=/build_cache/.gradle'
    - 'GRADLE_OPTS="-Dorg.gradle.daemon=false -Dkotlin.incremental=false"'
    - 'BRANCH_NAME=$BRANCH_NAME'
    - GCS_CACHE_BUCKET=gs://${PROJECT_ID}-cache/

- name: gcr.io/$POJECT_ID/restore_cache
  args:
  - --key=build-cache$( checksum build.gradle )
  <<: *env
```

oliverxchen:
@rharter: I agree with you on the `=` based arguments and having args from the same source. Those make it much cleaner to use.

Re: separate docker images – you're right, since it's just a different (later) script layer, separate docker images shouldn't add much to the build time. In which case this isn't a very important issue. In that case, the only advantage of having a single docker image is having the cache tool in a single "place". It would still be possible to have separate steps with a single docker image. Just use different entry points, or have a combined bash script and pass in a restore or save argument.

I'll have a chance to actually try these out in a few days so hopefully can have a more informed opinion...

nof20:
@kszalontai thanks a lot for your contribution.  Are you OK with the approach suggested above?  If we go with @rharter 's solution you can of course add features to it afterwards.

kszalontai:
@nof20 sorry for the late reply. sure i'm fine with @rharter 's solution!
i'm closing this.

PR #175, opened 2019-01-28
Adding Firebase test lab step as cloub build step 

RamzyHassan-VG:
by adding this step allows us to run UI test via firebase test lab. which lets you test your app on a range of devices and configurations, to give you a better idea of how it'll perform in your users' hands. To facilitate such broad testing, Test Lab runs your tests in batches called test matrices. In each matrix, your app is tested against multiple combinations of device configurations.


nof20:
Looks good.  Thanks!

PR #178, opened 2019-01-30
Test make example in main cloudbuild

Philmod:
I've set up a Google Cloud Build for this specific builder, as an example.

There is one caveat: I had to add a `dir` in the cloudbuild steps, which means people will have to run this specific builder from the root directory: `gcloud builds submit --config=make/cloudbuild.yaml make/.`

I added the run of the example in the make/cloudbuild.yaml.

### DIFF
--- a/make/cloudbuild.yaml
+++ b/make/cloudbuild.yaml
@@ -7,5 +7,12 @@ steps:
 - name: 'gcr.io/$PROJECT_ID/make'
   args: ['-v']
 
+# Test the examples with :latest
+
+# hello_world
+- name: 'gcr.io/$PROJECT_ID/make:latest'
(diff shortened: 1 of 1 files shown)


nof20:
@Philmod is this still necessary?  I'm a bit concerned about leaving things in an inconsistent state without much explanation.  Perhaps let's chat offline.

Philmod:
@nof20 The `dir` is necessary to work with the new triggers afaik. A better solution would be to define it at the trigger level, but I don't think it's possible.

codrienne:
Hey Mod. :) I'm going through the backlog of open PRs and am closing this PR for oldness. Feel free to re-open the PR or make a new one.

PR #186, opened 2019-02-09
Minimal docker image for slackbot

oliverxchen:
#### What this PR does

* Reduces the time to run the slackbot step in a build from 15-30 seconds to 4 seconds or less
* Reduces the size of the slackbot docker image from 371.5MB to 6.4MB (based on virtual sizes reported in GCR)

#### How it's done

Changed the `Dockerfile` to a two-stage build, with the second stage based on `alpine:latest`.

Reference: https://medium.com/@chemidy/create-the-smallest-and-secured-golang-docker-image-based-on-scratch-4752223b7324 

#### How it's tested

The new slackbot docker image was manually tested on the three examples and gives the same (but faster) results as the previous slackbot docker image.

PR #188, opened 2019-02-14
Adds save/restore cache builders.

rharter:
This PR adds two builders, one to save cache files and one to restore them.

### Key to identify cache files

The key allows builds to have multiple caches for things like dependencies, resources and incremental build artifacts.  Adding the `checksum` helper allows cache files to be easily updated when needs change meaning, when mixed with GCS object lifecycles, you're not paying for file storage for dependencies that are no longer needed.

### Remote (GCS) or local cache storage

Adding local caching allows this builder to be used with other cache storage mechanisms, like a company's internal resources.  The step can be used to store the cache in a local file, which can be sent by a following step to any place as needed.

### Step Versioning

While the current method of always updating the latest steps works when users are required to upload steps to their own GCR repository, it prevents the ability to have a shared public repository while still being able to update builders without breaking people builds.


rharter:
I wrote a blog post introducing these steps that highlight how easy, yet versatile, they were to use. 

https://ryanharter.com/blog/cloud-build-recipes-2/

rharter:
Based on feedback from @oliverxchen I've combined both of these builders into a single builder, since they're dependent on each other.  Two builders will still be built, to make use really simple, but only the entrypoint is different between them, so docker image downloading should be super fast.

oliverxchen:
That looks great, @rharter! I'll be able to test this out early next week.

nof20:
This is awesome @rharter , thanks a lot.  (The blog posts are great too!) 

@oliverxchen shout and let us know how you get on with testing.

oliverxchen:
It works very well! Thanks again, @rharter!

It's also possible to avoid building/pushing/storing the `restore_cache` and `save_cache` images in addition to the `cache` image. Instead, you can use the `cache` image with `entrypoint: 'restore_cache'` and `entrypoint: 'save_cache'`. I built the image with this https://github.com/rharter/cloud-builders-community/compare/rh/manage_cache...oliverxchen:manage_cache and used it in an actual build and it worked.

That's a cosmetic change though and won't reduce build times.

rharter:
That's certainly an option, but should be mlnoted that that makes the build step a little more complex (need to add the builder **and** the entrypoint to specify the action). The specific builders only add the entrypoint to the `cache` container so the added storage and download costs should be insignificant.

rharter:
That's certainly an option, but should be mlnoted that that makes the build step a little more complex (need to add the builder **and** the entrypoint to specify the action). The specific builders only add the entrypoint to the `cache` container so the added storage and download costs should be insignificant.

oliverxchen:
Ah, I was under the (wrong) impression that the "virtual size" reported in GCR was the amount of storage used in GCS. I looked in the GCS bucket directly and see that that's not the case.

So what you're saying makes sense: it's better to have separate builders so that the entrypoint isn't needed.

oliverxchen:
One thing I forgot to mention: I don't think the GCS restore and save cache examples in the readme will work properly. The path is indicated as `.cache/...`, so what's going to be cached is `/workspace/.cache/...`. But then in `restore_cache` the unpack command is: 
```
tar xpzf "$CACHE_FILE" -C /
```
so it will be restored to `/.cache/...` instead of `/workspace/.cache/...`.

I didn't actually test the example, but I ran into a related issue when I was using it in an actual build.

An easy fix would be to just change the readme to say that the path given in `save_cache` needs to be an absolute path and not a relative path.

rharter:
Good catch. I actually ran into that in one up my builds. 

The save command also has `-C /` so I thought that would do the trick for relative paths, but I was mistaken.

I'd love to get the command figured out so it could always do the right things for relative or absolute paths.

oliverxchen:
`-P` seems to work: https://github.com/rharter/cloud-builders-community/compare/rh/manage_cache...oliverxchen:manage_cache

rharter:
@oliverxchen @nof20 I haven't hit this state from @googlebot before.  Looks like this will take some manual intervention.

oliverxchen:
Strange, I'm using the same email address on this commit as for https://github.com/GoogleCloudPlatform/cloud-builders-community/pull/186. It's not automatically checking all commit authors?

@rharter, feel free to remove my commit and re-apply it yourself to get around this issue. Mine was a really small part of this.

nof20:
Hey, I checked and we have a signed CLA for @rharter but not for @oliverxchen.  Either Oliver needs to [sign the CLA here](https://cla.developers.google.com/clas), or Ryan needs to [squash that commit](https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git) (i.e. take responsibility for the content).  Thanks!

nof20:
Oh wait - I now see #186 .  Yeah that's weird, squashing the commits is the best solution.

rharter:
I read the issue a little differently, in that the CLA bot doesn't allow me to submit code on someone else's behalf without their consent. 

> 😕 The bad news is that it appears that one or more commits were authored or co-authored by someone other than the pull request submitter. We need to confirm that all authors are ok with their commits being contributed to this project. Please have them confirm that here in the pull request.
> 
> _Note to project maintainer: This is a terminal state, meaning the `cla/google` commit status will not change from this state. It's up to you to confirm consent of all the commit author(s), set the `cla` label to `yes` (if enabled on your project), and then merge this pull request when appropriate._

@nof20 Are you able to simply change the CLA label to `yes`? If not I'll squash.

nof20:
Ah yes - you're correct and I didn't read closely.  I read the comments above to mean @oliverxchen is fine with his code being included here, so I'll merge.  Thanks for everything!

jthegedus:
This is a great Builder! Thanks everyone for your work 🎉 I really wish the [`speeding-up-builds` docs here](https://cloud.google.com/cloud-build/docs/speeding-up-builds#caching_directories_with_google_cloud_storage) had mentioned this Builder. Truly awesome 😄

PR #194, opened 2019-02-24
Support for Swift build steps.

Andrewangeta:


Andrewangeta:
@nof20 Bot seems to be down also I've resigned the CLA.

Andrewangeta:
I signed it!

nof20:
This is awesome, thanks so much.  

Can you remove the `.DS_Store` files?  I also notice you're using the `drewskiwin:swift-latest` base image for latest rather than an official one, was there a reason for that?

On the CLA: I checked your details and it looks like `andrewangeta@gmail.com` is properly registered, but you have some commits here from `andrew.edwards@alliancedata.com` which is not.  Quickest way to fix this is to [squash your commits](https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git) using the gmail account - can you do that and let me know?

Andrewangeta:
@nof20 Done :)

nof20:
Perfect, thanks for sticking with it!

PR #196, opened 2019-02-25
Add support for dotenv files in "envsubst" builder

arikkfir:
This allows providing environment variables via dotenv files to the `envsubst` builder. 

This can be useful when the environment variables are dynamic in nature and need to be generated by another step first.


nof20:
Looks good, thanks!

PR #200, opened 2019-03-12
Fix /workspace is not within a known GOPATH/src

kevin-ip:
It works for me to address
```
Running: dep ensure -v
```

See also
https://github.com/GoogleCloudPlatform/cloud-builders-community/pull/69


kevin-ip:
Found a bug, fixing it

PR #201, opened 2019-03-12
Update README.md

YaguraStation:
Moving trailing instructions for permission of the cloudbuilder account to the front where the other roles are assigned. 

I understand why this was initially structured in this fashion, because the KMS API is enabled later in the doc. But I moved the KMS API enablement **also to the front** where the other APIs are enabled.


nof20:
Looks good.  Thanks

PR #202, opened 2019-03-12
Dep cloud builder fix

kevin-ip:
Second attempt to address
```
Running: dep ensure -v
```

See also
https://github.com/GoogleCloudPlatform/cloud-builders-community/pull/69


nof20:
@kevin-ip thanks a lot for doing this.

Can you add an example?  I'm thinking of it like a simple unit test, which we can use to demonstrate that this fixes problems.

kevin-ip:
@nof20 thanks for reviewing. 

I just added a simple unit test. Could you help take a look when you get a chance? Thanks

nof20:
Looks good.  Thanks

PR #203, opened 2019-03-13
add s2i support to google cloud build

yxue-kabam:
This will create a s2i builder image that can be pushed to `gcr.io/$PROJECT_ID/s2i`


nof20:
Looks good, thanks!

PR #206, opened 2019-03-15
add shellcheck builder

n0npax:
This PR providing shellcheck builder.
Shellcheck is nice tool which provides linting for shell scripts.


nof20:
Thanks so much for submitting this.  Should your example output something, like a list of issues with the example shell-script?  At the moment it just runs and completes.

n0npax:
Hey, by default provided example script has no linting issues, so output should be empty and run should be successful. If you want to see a issues list please uncomment 2nd `tr` in the provided example script (I added a comment there in the midtime)

nof20:
Looks great.  Thanks!

PR #213, opened 2019-03-27
adds an option to save_cache to skip saving if cache exists in GCS

oliverxchen:
**Problem this PR solves:** Certain types of builds will not change the cache, but `save_cache` will still compress the cache and re-upload it to GCS. This PR provides an option (`-s` or `--skip_if_exists`) to check if the cache file exists in GCS and skips the compress/upload steps if it does exist.

For example, in my node based project I'm caching the `node_modules` directory. If it can be restored from cache, I don't need to run `npm ci` and the `node_modules` directory won't change at all. That directory can be hundreds of megabytes, so the unnecessary compress/upload steps can add up to 30 seconds to the build.

@rharter, wanted to get your feedback on this. If you think it makes sense I can also describe this in the readme file. And any suggestions you have to improve this would be appreciated!

### DIFF
--- a/cache/save_cache
+++ b/cache/save_cache
@@ -11,12 +11,12 @@ function print_usage {
   echo
   echo "Saves the specified paths to a cache file located in the out directory."
   echo
-  echo "  -b, --bucket          The cloud storage bucket to upload the cache to. [optional]"
-  echo "  -o, --out             The output directory to write the cache to. [optional]"
-  echo "  -k, --key             The cache key used for this cache file."
-  echo "  -p, --path            The files to store in the cache. Can be repeated."
-  echo "  -t, --threshold       The parallel composite upload threshold [default: 50M]"
-  echo "  -s, --skip_if_exists  Skips the save if the cache file already exists in GCS."
+  echo "  -b, --bucket      The cloud storage bucket to upload the cache to. [optional]"
+  echo "  -o, --out         The output directory to write the cache to. [optional]"
+  echo "  -k, --key         The cache key used for this cache file."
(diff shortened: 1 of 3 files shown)


cache/save_cache
```diff
@@ -11,12 +11,13 @@ function print_usage {
...
-  echo "  -k, --key        The cache key used for this cache file."
-  echo "  -p, --path       The files to store in the cache. Can be repeated."
-  echo "  -t, --threshold  The parallel composite upload threshold [default: 50M]"
-  echo 
+  echo "  -b, --bucket          The cloud storage bucket to upload the cache to. [optional]"
+  echo "  -o, --out             The output directory to write the cache to. [optional]"
+  echo "  -k, --key             The cache key used for this cache file."
+  echo "  -p, --path            The files to store in the cache. Can be repeated."
+  echo "  -t, --threshold       The parallel composite upload threshold [default: 50M]"
+  echo "  -s, --skip_if_exists  Skips the save if the cache file already exists in GCS."
```
rharter: I think `-` is a more common word separator in command line apps, as opposed to `_`, which would make this `--skip-if-exists`.

Also, for consistency with other bash commands would you be opposed to using something like `-n, --no-clobber` from [`cp`](http://manpages.ubuntu.com/manpages/precise/en/man1/cp.1.html) and [`mv`](http://manpages.ubuntu.com/manpages/precise/en/man1/mv.1.html)?

oliverxchen:
Thanks, @rharter! Good suggestions, I've implemented them and added an explanation to the readme.

PR #215, opened 2019-04-03
Double quote $@ to ensure arguments are forwarded properly

Nilhcem:
`for_branch` is sometimes useful (_it would be great to have native branch filtering per steps, though_).
However, when passing arguments surrounded with quotes that contain spaces, the `$@` splits those arguments per space, which is not ideal.
For instance, the following curl script that sends an APK to HockeyApp only when merging to master would not work:

```
for_branch master curl -H "X-HockeyAppToken: $_TOKEN" $URL
```

We expect:
```
curl
-H
"X-HockeyAppToken: $_TOKEN"
$URL
```

and not:
```
curl
-H
X-HockeyAppToken:
$_TOKEN
$URL
```

This commit adds double quotes to `$@` to ensure arguments are forwarded properly.

### DIFF
--- a/android/for_branch
+++ b/android/for_branch
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Runs the arguments as if they were passed directly to bash if the 
+# Runs the arguments as if they were passed directly to bash if the
 # BRANCH_NAME environment variable matches the branch argument in the
 # first position.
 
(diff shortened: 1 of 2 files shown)


nof20:
@rharter could you take a look at this?  Seems sensible to me.

nof20:
Thanks!

PR #217, opened 2019-04-05
Add a builder for Pulumi

praneetloke:
This is a simple builder for [Pulumi](https://www.pulumi.com).


praneetloke:
I signed it!

pulumi/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+```
+env: ["PULUMI_ACCESS_TOKEN=$_PULUMI_ACCESS_TOKEN"]
+```
+
+## Example: Run a preview on a stack
+
+[![Deploy](https://get.pulumi.com/new/button.svg)](https://app.pulumi.com/new)
+
+**Be sure to set `_PULUMI_ACCESS_TOKEN` in the build trigger substitutions. The leading `_` is important when you define the substitution.**
```
ImJasonH: Access tokens should be passed as secrets, or else they're not treated as sensitive in our backend. Any user with access to describe builds would be able to see this token, for instance.

Secrets are encrypted with a KMS key and only users with decrypt access can see them, and even then only if they explicitly decrypt the token.

praneetloke:
Thank you for the review, @ImJasonH. I'll post the build log as proof tomorrow.

pulumi/README.md
praneetloke: Understood. Since this is just a sample of how a user _could_ pass the token, would it be ok if I reference the kms encryption for the build variable and the usage of `secretEnv`? This example is meant to be a quick start. I do have the encrypted build variables referenced in a documentation PR opened with Pulumi [here](https://github.com/pulumi/docs/pull/974).

praneetloke:
Here's a build log for a sample repo I have on Cloud Source. The project deploys a Cloud Function with an HTTP trigger. It uses the `pulumi-node` builder from this repo as one of the build steps. Truncated some build output for brevity.

```
starting build "b207ed33-cb91-4963-ac48-ebbb56931752"

FETCHSOURCE
Initialized empty Git repository in /workspace/.git/
From https://source.developers.google.com/p/cloud-build-samples/r/pulumi-gcp-cloud-functions
 * branch            24888d97917d3faeb2feb46f182826201881ad4f -> FETCH_HEAD
HEAD is now at 24888d9 Run yarn install.
BUILD
Pulling image: gcr.io/cloud-build-samples/pulumi-node
Using default tag: latest
latest: Pulling from cloud-build-samples/pulumi-node
e79bb959ec00: Already exists
d4b7902036fe: Already exists
....
....
dfea7bf35707: Waiting
c442a354fd24: Verifying Checksum
c442a354fd24: Download complete
....
....
dfea7bf35707: Pull complete
Digest: sha256:4a8212cdf0a904d0114584155a0be5f58d37b3a9d22440dc916238a3949bf472
Status: Downloaded newer image for gcr.io/cloud-build-samples/pulumi-node:latest
+ yarn install
yarn install v1.12.3
warning package.json: No license field
warning gcp-ts-functions@1.0.0: No license field
....
....
....
+ case $BUILD_TYPE in
+ pulumi preview
Previewing update (dev):

    pulumi:pulumi:Stack gcp-functions-dev running 
    gcp:cloudfunctions:CallbackFunction greeting  
    gcp:storage:Bucket greeting  
    gcp:storage:BucketObject greeting  
    gcp:cloudfunctions:Function greeting  
 
Resources:
    5 unchanged
....
PUSH
DONE
```

pulumi/README.md
ImJasonH: I'd like to very strongly recommend users not pass their tokens that way. Users have a tendency to copy/paste samples to get up and running, so it's important to either have the sample do the right thing, or if it's necessary that it do the wrong thing, call that out very very loudly as a wrong thing. If you do end up having the sample use a substitution, naming the substitution `_SUBSTITUTION_FOR_TESTING_INSECURE_USE_SECRET_INSTEAD` or something similarly scary sounding.

praneetloke: Just seeing your latest comment. I have updated the doc with information about encrypting env vars. I like the idea of making the substitution variable explicitly call out this is just for an example. I'll make that change as well.

praneetloke: Done.

praneetloke:
Ping.

pulumi/README.md
```diff
@@ -0,0 +1,75 @@
...
+$ cd my-project
+$ gcloud builds submit .
+```
+
+At this point you can automate the build using triggers via GCP Repos or Github. See [this](https://pulumi.io/reference/cd-google-cloud-build.html) doc on how to integrate Pulumi into your CI/CD setup.
+
+## Authentication
+
+The access token required for a non-interactive login is passed in via an environment variable.
+Instead of hardcoding the value of the access token in your build configuration. You can use [substitutions](https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values).
```
ImJasonH: This should recommend using a secret env var, even though the example uses a substitution for ease.

Using a substitution is effectively identical to hard-coding the value in your build config, since any user with permission can see the substitution value using `gcloud builds describe` or the Cloud Console.

Above, you've changed the substitution to declare that it's `INSECURE`, but it would be helpful to explain why it's insecure, and how to do something more secure using secrets.

praneetloke: I have removed this, as well as the separate Examples section. Since the quickstart steps walk through the same build configuration sample, I have moved the information block about encrypted secrets to that section.

praneetloke: Ping.

ImJasonH:
Sorry for the delay! LGTM

niedbalski:
This break builds :-) everywhere on GCB.

praneetloke:
Hmm.. that's strange. @niedbalski can you please simply revert the git submodule file in your commit?

praneetloke:
It is this file: https://github.com/GoogleCloudPlatform/cloud-builders-community/blob/master/.gitmodules

praneetloke:
@niedbalski can you please share the cloudbuild config you used that caused the issue for you? I'd like to repro this myself.

PR #219, opened 2019-04-08
Add a custom build step for singularity

schoonovernumerics:
This patch provides a Dockerfile, cloudbuild.yaml, and an example for adding singularity as a build step for gcloud builds.


nof20:
Looks good.  Thanks!

PR #220, opened 2019-04-08
update kustomize version to current stable v2.0.3

murog:
Previous Dockerfile was a major version behind
[v2.0.0 release notes](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/version2.0.0.md)


nof20:
Whoops.  I missed this, but came to use Kustomize recently and upgraded it myself, so the effect is the same 😄 

Thanks though!

PR #227, opened 2019-04-30
Update Hugo to 0.55.0

polleyg:
Update Hugo to 0.55.0


polleyg:
Some Hugo themes only support >=0.55.0 e.g:

Screen Shot 2019-04-30 at 12 45 04 pm

nof20:
Looks good.  Thanks!

PR #229, opened 2019-05-08
Base Image Builder

johnlabarge:
This uses something like remote builder to build a base-image-builder for GCP.


nof20:
Looks good.  Thank you!

PR #231, opened 2019-05-08
container-diff [issue 230] updates for Bazel

SpreetailHurley:


SpreetailHurley:
updated email address in CLA

ImJasonH:
There's no real reason for container-diff to be built with Bazel at all, and it causes breakages like this sometimes. We should consider removing Bazel from the picture and just building the image with a Dockerfile that fetches the binary and puts it in an image.

SpreetailHurley:
Hi Jason,
Would there be value in going the other direction - using Bazel to instead base container-diff on distroless?
Thanks.

PR #232, opened 2019-05-09
Fix builds command in dep/README.md

tommy-sho:
Fix command in dep/README.md to use `gcloud builds` command


nof20:
Looks good.  Thanks!

PR #235, opened 2019-05-16
set permissions to skaffold bash

simonz130:
When I tried to execute skaffold custom builder, it failed with permissions not set on skaffold.bash.


nof20:
Looks good.  Thanks

PR #237, opened 2019-05-22
Swap ubuntu for alpine for packer image build

mcrmonkey:
Avoids having to install additional packages to perform some basic tasks
Alpine pinned to 3.9


nof20:
Looks good, thanks.

PR #245, opened 2019-06-14
adds unzip

rovolution:
Adds image with [`unzip`](https://linux.die.net/man/1/unzip)


nof20:
Thanks for your submission!  I just merged #247 which provides zip and unzip, I think that covers everything.

PR #247, opened 2019-06-18
Add zip builder

etrepum:
This adds a zip builder, modeled after the tar builder that is already present.

The use case I have is to package up a Google Cloud Function and upload it to an artifact bucket for later deployment (manually at this time).


nof20:
Looks good, thanks!

PR #248, opened 2019-06-18
Implemented support for using custom Windows images

hiranya911:
Firebase team would like to use this builder to package some of our .NET libraries. For that we want the ability to use our own Windows images with .NET Framework and .NET Core pre-installed. This PR makes it possible to specify a custom image for the builder.

I also noticed that Windows images often take longer than 2 minutes to start. So I'm increasing the timeout to 5 minutes as well.


nof20:
Looks good, thank you!

PR #249, opened 2019-06-20
Add azure cli and aks kubectl builders

scotth1:
Cloud Build steps for az and az-kubectl.

### DIFF
--- a/az-kubectl/Dockerfile
+++ b/az-kubectl/Dockerfile
@@ -1,4 +1,4 @@
-FROM gcr.io/learning-resources-hub/az
+FROM gcr.io/$PROJECT_ID/az
 
 # Install kubectl component
 RUN /usr/local/bin/az aks install-cli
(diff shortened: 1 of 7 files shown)


edu-shamilton1:
I can confirm consent.
I was an idiot and forgot to change user.email before committing code from a different account by accident.

az-kubectl/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM gcr.io/learning-resources-hub/az
```
nof20: Can you explain who owns this project/container and how it's maintained?  It's opaque here & we're inviting people to build dependencies on it.

Is there an official image maintained by MS on MCR (like the Azure example below)?

scotth1: Ah, that should be in the same project, there are two steps built, first one is $PROJECT_ID/az and the second one is $PROJECT_ID/az-kubectl which is built FROM the first.
Looks like I forgot undo all the project specific things, I'll make a new commit.

nof20:
Looks good.  Thanks!

PR #250, opened 2019-06-22
Add builder for Esy

tmattio:
Esy is a project manager for ReasonML and OCaml.

This PR adds a builder for Esy that can be used like this:
```
steps:
- name: 'gcr.io/project-id/esy'
  args: ['install']
- name: 'gcr.io/project-id/esy'
  args: ['build']
```


nof20:
Looks good, thanks!

PR #254, opened 2019-07-03
Add Node.js HTTP Cloud Function Cloud Builder

charles-salmon:
This PR adds a cloud builder which builds, tests and deploys a Cloud Function that is written using Node.js and intended to respond to an HTTP trigger.


charles-salmon:
I'm going to abandon this PR for now. I built this for a specific use case, but I think it would make sense for a community builder to be more general.

The approach I intend to take is as follows:

* Generalise image such that all supported runtimes (`nodejs6`, `nodejs8`, `nodejs10`, `python37` and `go111`) can be built using the same image. I'll use `nvm` to manage node versions, and change the base image to `google/cloud-sdk:slim`.
* Use the Cobra framework to write the CLI, such that this can be covered in unit tests.
* Modify the CLI to pass all flags/parameters to `gcloud functions deploy`, except pre-deploy parameters (e.g. `--yarn-scripts` / `--npm-scripts`) which will define what should happen prior to deployment (and assume sensible defaults).

PR #260, opened 2019-07-10
[protoc] Parameterize protoc version|architecture

DazWilkin:
These changes should:

+ Update the protoc version from `3.6.1` to the current release `3.8.0`.
+ Provide a way to more easily build current version|architectures as needed.

To build for a different architecture and version:
```bash
gcloud builds submit . \
--config=cloudbuild.yaml \
--substitutions=_VERS=${VERS},_ARCH=${ARCH} \
--project=${PROJECT}
```

However, one challenge I'm having using this Builder is that `protoc` often depends upon external binaries, e.g. `protoc-gen-go` to produce language (Golang) specific sources. I've not yet worked out how to achieve this using this container image.

### DIFF
--- a/protoc/Dockerfile
+++ b/protoc/Dockerfile
@@ -1,16 +1,18 @@
-FROM ubuntu
+FROM ubuntu:19.10
 
-ARG PROTOC_VERSION=3.6.1
-ARG PROTOC_TARGET=linux-x86_64
-ARG ASSET_NAME=protoc-${PROTOC_VERSION}-${PROTOC_TARGET}.zip
+ARG VERS=3.8.0
+ARG ARCH=linux-x86_64
 
-RUN apt-get -qy update && apt-get -qy install python wget unzip && rm -rf /var/lib/apt/lists/*
+RUN apt-get -qy update && \
+    apt-get -qy install wget unzip && \
+    rm -rf /var/lib/apt/lists/*
 
-RUN echo "${PROTOC_VERSION}/${ASSET_NAME}"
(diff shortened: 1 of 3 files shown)


DazWilkin:
I discovered the [solution](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.compiler.plugin) to the compiler plugin problem I was facing.

Now I can `go get` e.g. `protoc-gen-go` and `protoc-gen-grpc-gateway` using the following approach and without having to munge `${PATH}`:
```bash
- name: gcr.io/${PROJECT_ID}/protoc
  args:
  - --proto_path=...
  - --proto_path=...
  - --plugin=protoc-gen-go=/path/to/protoc-gen-go
  - --plugin=protoc-gen-grpc-gateway=/path/to/protoc-gen-grpc-gateway
  - --go_out=plugins=grpc:...
  - --grpc-gateway_out=logtostderr=true:...
  - ...
```
This may be worth adding to the `protoc` Builder's documentation?

protoc/Dockerfile
```diff
@@ -1,16 +1,18 @@
-FROM ubuntu
+FROM ubuntu:19.10
```
ImJasonH: Does it need 19.10?

DazWilkin: I think it's better to be more precise than the vagaries of the alternative of `:latest`.

It would be still if I were to use a specific image digest.

Thoughts?

ImJasonH: I'm for it, I was just curious.

PR #265, opened 2019-07-14
Fix caching volume issues in Android examples

pavlospt:
There was a wrong path referenced in one sample and on the other the indentation was wrong.

### DIFF
--- a/android/examples/build-test-deploy.yaml
+++ b/android/examples/build-test-deploy.yaml
@@ -27,7 +27,7 @@ steps:
   args: 
   - '-c'
   - |
-    cat /config/buildnum | awk '{print $1"+1"}' | bc | tee /config/buildnum | awk '{print "BUILD_NUM="$1 }' | tee .buildnum
+    cat /config/buildnum | awk '{$1=1+$1;print}' | tee /config/buildnum | awk '{print "BUILD_NUM="$1}'
   waitFor: ['copy_config']
   volumes:
   - name: 'config'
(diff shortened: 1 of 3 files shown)


pavlospt:
Hey @rharter the documentation of Cloud build though suggests that indentation, so that is why I kept it consistent between volume declarations and not among the general indentation of the file! I could revert if it's an issue!

rharter:
I think it makes sense that the examples use the same indentation as the documentation.  I would argue, however, that consistency within files is more important.

With that in mind, I'm not opposed to changing the list indentation style of the file, I just think that if we're going to update the indentation of some lists to match the docs we should update them all.  I would vote for adding the rest of the indentation fixes to this PR.

pavlospt:
Sure that sounds fine! I will update the PR accordingly!

pavlospt:
@rharter I updated the syntax for `waitFor` and fixed any indentation issues I spotted! Let me know if it is ok?

rharter:
@nof20 would love to see this one merged, as well.  Thanks!

pavlospt:
@sanastos would it be possible to take a look at this one as well?

leeonlee:
Looks good to me, thanks!

FYI: Cloud Build now supports global volumes so they only have to be defined once at the build level and they'll be mounted to each step: https://cloud.google.com/cloud-build/docs/api/reference/rest/Shared.Types/Build#BuildOptions

pavlospt:
@leeonlee Great, thanks for sharing this!

PR #267, opened 2019-07-16
Fix tar cloudbuild.yaml

pavlospt:
Tar was failing to be deployed since the `Dockerfile` could not be located when building the image. This was because the whole `source.tgz` was downloaded and the Dockerfile mentioned in the `cloudbuild.yaml` was not referenced based on the `tar` folder path.

### DIFF
--- a/tar/cloudbuild.yaml
+++ b/tar/cloudbuild.yaml
@@ -1,6 +1,10 @@
 steps:
 - name: gcr.io/cloud-builders/docker
-  args: ['build', '-t', 'gcr.io/$PROJECT_ID/tar', '.']
+  args:
+  - 'build'
+  - '--tag=gcr.io/$PROJECT_ID/tar'
+  - '--file=tar/Dockerfile'
+  - '.'
 
 images: ['gcr.io/$PROJECT_ID/tar']
 tags: ['cloud-builders-community']


rharter:
Hmm, something looks a little fishy about this one. The Dockerfile should be relative to the cloud build config file, so you shouldn't need to add the `tar/` prefix. Are you running cloud build from a different directory.

pavlospt:
@rharter Not sure wha you mean by that :/ I just tried deploying that to a project and it failed to be deployed! This was not the case though for the Android image, which was correctly deployed by following the same approach! All I did was just: `gcloud builds submit --config=cloudbuild.yaml` by using the CloudBuild YAMLs from `android` & `tar` folders respectively.

rharter:
Yeah, I was curious if you might've done `gcloud builds submit --config=tar/cloudbuild.yaml` from the root directory. I expect the docker directive to be relative to the cloudbuild.yaml file (or the current directory, at the least). 

That's how the [android](https://github.com/GoogleCloudPlatform/cloud-builders-community/blob/master/android/cloudbuild.yaml#L7) code, and others, do it. No need for parent dir to be explicitly included.

rharter:
Confirmed, when the step runs, it zips up your current directory, then creates a docker image based on the `gcr.io/cloud-builders/docker` image, explodes the zip from before in `/workspace`, and `/workspace` is set to the current directory (almost, it's not exploded for each step, but that's the effect), and then uses the default dockerfile, `.Dockerfile`, since it hasn't been overridden.

So as long as you run `gcloud builds submit --config=cloudbuild.yaml` from the `/tar` directory, it should all work as expected.  I'd expect to see the issue you saw if you ran `gcloud builds submit --config=tar/cloudbuild.yaml` from the parent of the tar directory (root of the repo).  Could that be what's happened?

pavlospt:
@rharter yeah indeed, that is what happened, since I did the same when deploying the `android` image as well! Maybe that should be documented somehow, could prove useful for other people in the future as well! Thanks for investigating it!

PR #273, opened 2019-07-26
Updated JMeter version to 5.1. 

rprakashg:
Updated JMeter version to 5.1 as 5.0 binaries are no longer available to download


nof20:
Looks good, thanks!

PR #278, opened 2019-08-06
Fix incorrect example cloudbuild.yaml for binauthz-attestations

wcalandro:
The existing cloudbuild.yaml had an incorrect entrypoint for some of the steps, so it wasn't valid.


leeonlee:
Appreciate the contribution!

PR #279, opened 2019-08-07
Make it possible to specify an arbitrary Terraform version

iangelov:
In our environment, we found it useful to have the ability to build container images with arbitrary Terraform versions. To that end, I made the following changes which should allow one to do so.

I also added "-Ss --fail" to the curl command line, so that curl doesn't output progress if everything is working correctly, but do so if there's an error, and fail if necessary. This should produce less garbage output and streamline diagnosing issues. Additionally, the container image is properly tagged with the Terraform version, which allows one to be more explicit about which version should actually be used depending on the use case.

The current implementation allows for the Dockerfile to be used independently from Cloud Build as well.

### DIFF
--- a/terraform/Dockerfile
+++ b/terraform/Dockerfile
@@ -1,7 +1,7 @@
 FROM gcr.io/cloud-builders/gcloud
 
-ARG TERRAFORM_VERSION=0.12.6
-ARG TERRAFORM_VERSION_SHA256SUM=6544eb55b3e916affeea0a46fe785329c36de1ba1bdb51ca5239d3567101876f
+ARG TERRAFORM_VERSION
+ARG TERRAFORM_VERSION_SHA256SUM
 
 RUN apt-get update && \
     apt-get -y install curl unzip ca-certificates && \
--- a/terraform/Dockerfile
+++ b/terraform/Dockerfile
@@ -1,11 +1,11 @@
 FROM gcr.io/cloud-builders/gcloud
 
-ENV TERRAFORM_VERSION=0.12.6
-ENV TERRAFORM_VERSION_SHA256SUM=6544eb55b3e916affeea0a46fe785329c36de1ba1bdb51ca5239d3567101876f
+ARG TERRAFORM_VERSION=0.12.6
+ARG TERRAFORM_VERSION_SHA256SUM=6544eb55b3e916affeea0a46fe785329c36de1ba1bdb51ca5239d3567101876f
 
 RUN apt-get update && \
     apt-get -y install curl unzip ca-certificates && \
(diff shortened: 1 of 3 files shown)


terraform/Dockerfile
```diff
@@ -1,11 +1,11 @@
 FROM gcr.io/cloud-builders/gcloud
 
-ENV TERRAFORM_VERSION=0.12.6
-ENV TERRAFORM_VERSION_SHA256SUM=6544eb55b3e916affeea0a46fe785329c36de1ba1bdb51ca5239d3567101876f
+ARG TERRAFORM_VERSION=0.12.6
+ARG TERRAFORM_VERSION_SHA256SUM=6544eb55b3e916affeea0a46fe785329c36de1ba1bdb51ca5239d3567101876f
```
leeonlee: Removing the default args gives us only one place to manage the default version in the substitutions

terraform/cloudbuild.yaml
```diff
@@ -1,9 +1,21 @@
...
+  - --build-arg
+  - TERRAFORM_VERSION
+  - --build-arg
+  - TERRAFORM_VERSION_SHA256SUM
+  - --tag=gcr.io/${PROJECT_ID}/terraform:${_TERRAFORM_VERSION}
+  - .
+- name: 'gcr.io/${PROJECT_ID}/terraform:${_TERRAFORM_VERSION}'
   args: ['version']
-images: ['gcr.io/${PROJECT_ID}/terraform']
-tags: ['cloud-builders-community']
```
leeonlee: Why was the tag removed?

iangelov: Are you referring to the implicit value of the tag "latest", or to the cloudbuild.yaml tag? I consider the former to be a good practice in general, although not necessarily what's intended in this case. The latter didn't exist at the time of the fork, but I've made sure to add it as it's a great thing to filter on. In any case - let's clarify the intent and I will make the necessary changes.

terraform/Dockerfile
iangelov: I left it in place to facilitate local testing, but it makes sense. I've removed the hardcoded values in the Dockerfile.

iangelov:
@googlebot I fixed it.

terraform/cloudbuild.yaml
leeonlee: Sorry I wasn't clear, I was referring to the cloud build tag.

leeonlee:
The changes look good to me, just need to fix the CLA check.

leeonlee:
Thanks!

iangelov:
@leeonlee - thank you for reviewing and merging this pull request!

PR #281, opened 2019-08-15
[protoc] fixes: documentation and Dockerfile

DazWilkin:
Fixed a couple of issues with the `protoc` builder:

+ Corrected reference to user-defined substitutions
+ Added caveat wrt the builder failing to build with `protoc` release candidates
+ Removed incorrect static tag from images when built.


DazWilkin:
@googlebot I signed it!

sanastos:
Thanks for the fix!

PR #282, opened 2019-08-16
Swap ubuntu for alpine for terraform image build

mcrmonkey:
Reduces need for installation of additional packages to perform basic
tasks during build time.

Same image build process as packer. alpine pinned to 3.9.


leeonlee:
Hi, there was a recent merge that has some overlapping changes (using build_args to specify terraform version): https://github.com/GoogleCloudPlatform/cloud-builders-community/pull/279

Please resolve conflicts and I'll take a look, thanks!

mcrmonkey:
Conflicts, I believe, are now corrected...

I've also adjusted the `build-arg` options so that they aren't just comping out as `--build-arg 0.12.8` and instead are `--build-args TERRAFORM_VERSION=0.12.8`

terraform/Dockerfile
```diff
@@ -1,21 +1,17 @@
...
-    rm -rf /var/lib/apt/lists/*
+
+COPY terraform_${TERRAFORM_VERSION}_linux_amd64.zip terraform_${TERRAFORM_VERSION}_linux_amd64.zip
+RUN echo "${TERRAFORM_VERSION_SHA256SUM}  terraform_${TERRAFORM_VERSION}_linux_amd64.zip" > checksum && sha256sum -c checksum
+RUN unzip terraform_${TERRAFORM_VERSION}_linux_amd64.zip
+
+
+FROM gcr.io/cloud-builders/gcloud
 
 ENV PATH=/builder/terraform/:$PATH
```
leeonlee: Now that we're copying `terraform` to `/usr/bin/terraform`, this line is unneeded. I have no strong opinions on `/builder/terraform` vs `/usr/bin/terraform`.

terraform/cloudbuild.yaml
```diff
@@ -4,16 +4,18 @@ substitutions:
...
+  args: ["https://releases.hashicorp.com/terraform/${_TERRAFORM_VERSION}/terraform_${_TERRAFORM_VERSION}_linux_amd64.zip"]
 - name: 'gcr.io/cloud-builders/docker'
   env:
   - 'TERRAFORM_VERSION=${_TERRAFORM_VERSION}'
   - 'TERRAFORM_VERSION_SHA256SUM=${_TERRAFORM_VERSION_SHA256SUM}'
   args:
   - build
   - --build-arg
-  - TERRAFORM_VERSION
+  - TERRAFORM_VERSION=${_TERRAFORM_VERSION}
```
leeonlee: > I've also adjusted the `build-arg` options so that they aren't just comping out as `--build-arg 0.12.8` and instead are `--build-args TERRAFORM_VERSION=0.12.8`

I'm under the impression that the build-arg was not incorrect but I have no strong opinion here either: 

"You may also use the --build-arg flag without a value, in which case the value from the local environment will be propagated into the Docker container being built" [link](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg)

terraform/Dockerfile
mcrmonkey: I've removed the `ENV` line but looks like the `RUN` and `FROM` lines are highlighted here :thinking:

leeonlee: Yeah, the editor grabbed an extra couple of lines 🤷‍♀️

leeonlee:
Thanks for the contribution!

PR #285, opened 2019-08-26
[cache] Introduce key_fallback option on `cache#restore_cache` builder

pavlospt:
# Background
Large projects (like ours for example) utilizing the `cache` builder, can not afford a lot of cache misses in case of a single change to a file modifying the actual cache checksum. Other CI/CD solutions (e.g.: CircleCI) that support out-of-the-box cache save & restoration, support multiple key patterns for cache restoration in case the key is not found.

# Contents of the PR
This PR modifies `restore_cache` script to accept a `key_fallback` option that will be used in case there is a cache miss from the `key` option. 

It also refactors the way `restore_cache` was working in order to rely on file existence checks instead of failed file copy attempt to attest a cache file's existence.

# Usage
Specify `key_fallback` option in `restore_cache` builder in order to use a fallback name pattern. 

Example
```yaml
- name: gcr.io/$PROJECT_ID/restore_cache
  id: restore_cache
  waitFor: ['generate_cache_checksum']
  args: [
    '--bucket=gs://${_CACHE_BUCKET}',
    '--key=gradle-$( checksum checksum.txt )',
    '--key_fallback=gradle-',
  ]
```

### DIFF
--- a/cache/README.md
+++ b/cache/README.md
@@ -122,3 +122,15 @@ This `cloudbuild.yaml` restores the files from the compressed cache file identif
   - name: 'cache'
     path: '/cache'
 ```
+
+### Restore a cache with a fallback key
+
+```yaml
+- name: gcr.io/$PROJECT_ID/restore_cache
+  id: restore_cache
+  args: [
+    '--bucket=gs://${_CACHE_BUCKET}',
+    '--key=gradle-$( checksum checksum.txt )',
+    '--key_fallback=gradle-',
+  ]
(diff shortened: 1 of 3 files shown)


sanastos:
Looks good, but would you mind putting your usage example from the PR description into the README.md examples section? I think it would be helpful.

pavlospt:
@sanastos done!

sanastos:
Thanks!

PR #287, opened 2019-08-28
[envsubst] add an option for SHELL-FORMAT & set globstar on

reppets:
some improvement for the envsubst builder:
* Added `-s` option to specify a SHELL-FORMAT string to `envsubst` command.
* Set globstar on for more flexible file-matching

### DIFF
--- a/envsubst/README.markdown
+++ b/envsubst/README.markdown
@@ -29,12 +29,12 @@ This builder can also accept wildcards! Lets say you have another file called `i
 The planet ${PLANET} is the next one on the solar system!
 ```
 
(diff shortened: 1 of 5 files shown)


sanastos:
The shell-format change seems incontrovertibly a clear improvement to me, but the globstar change modifies behavior that people could theoretically be depending on. Can you gate that behind a flag as well?

If you prefer to split this into two PRs I'll merge the shell-format one immediately.

sanastos:
OK, that's a fair point. Thanks for your contribution!

PR #290, opened 2019-09-03
Support regex match for branch name

aniketawati:
Certain build/test steps would need to be scoped on a branch pattern or multiple branches. So supporting regex in branch name would help configure steps easily. 
For example, if you want create release builds only on master and dev branches, and configure debug builds for feature branches, regex branch pattern would help.

`- args: ['for_branch', '(master|dev)',...]`
`- args: ['for_branch', '(master|dev|feature\/.*)',...]`

### DIFF
--- a/android/README.md
+++ b/android/README.md
@@ -15,9 +15,20 @@ export BUILD_NUM=123
 
 ## Run Only On Specific Branch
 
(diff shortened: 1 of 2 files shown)


aniketawati:
@googlebot I signed it!

sanastos:
This looks good, but would you mind updating the documentation to reflect that this takes a regex now?

Just a quick note [here](https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/android#run-only-on-specific-branch) would be sufficient.

sanastos:
Thanks for your contribution!

PR #293, opened 2019-09-07
Adapting to Helm v3

markmssd:
Helm is now hosted there, https://helm.sh/blog/get-helm-sh/

~Since it's the case since v2.0.0-alpha.5, it is not a breaking change :)~

I guess since Helm v2 also had `tiller` and its folder was also being stripped... that won't work anymore.... maybe it'd be wise to only merge this once v3 is out of beta?

### DIFF
--- a/helm/Dockerfile
+++ b/helm/Dockerfile
@@ -1,6 +1,6 @@
 FROM gcr.io/cloud-builders/gcloud
 
-ARG HELM_VERSION=v2.14.0
+ARG HELM_VERSION=v2.14.3
 
 COPY helm.bash /builder/helm.bash
 
--- a/helm/Dockerfile
+++ b/helm/Dockerfile
@@ -1,6 +1,6 @@
 FROM gcr.io/cloud-builders/gcloud
 
-ARG HELM_VERSION=v3.0.0-beta.3
+ARG HELM_VERSION=v2.14.0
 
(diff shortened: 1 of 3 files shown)


rimusz:
it would make more sense to keep helm v2, as helm v3 is only beta, but having both binaries it will allow to test releases with helm v3 as well

markmssd:
@rimusz I've reverted to Helm v2.

If I can find a way to strip `linux-amd64/tiller` only if it exists, that'd be good to merge... would you have suggestions?

rimusz:
tiller setup is fine there, just please use the latest helm v2 14.3

sanastos:
After update to 14.3 per @rimusz request I will merge, thanks!

markmssd:
That’s already done. However I still need to make more adjustments, e.g. to not call “helm init —client-only” when using v3, because that command is dropped.

It will do those changes soon

markmssd:
Changes done! Sorry for this long time.

I've modified `helm.bash` so it runs `helm init --client-only` only if using Helm v2. This command doesn't exist in Helm v3.

Let me know if there's anything missing to have it merged!

markmssd:
@googlebot I fixed it.

sanastos:
Looks good, thanks for your contribution!

PR #294, opened 2019-09-07
Add Zola as build command

shuluster:
I want to use static site generator [Zola](https://www.getzola.org/), similar to Hugo. They are so similar that I mostly copied from the [cloud builder for Hugo](https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/hugo).


shuluster:
@googlebot I signed it!

sanastos:
Looks good, thanks for your contribution!

PR #295, opened 2019-09-09
bump Skaffold version to v0.37.1

balopat:
Context:https://github.com/GoogleContainerTools/skaffold/issues/2837


sanastos:
The description says 'bump' but this isn't exactly a bump so much as a policy change right? And at the moment they should be functionally equivalent? 

In the context you mention that you're not sure that latest is the right thing to use but don't really expand on that. Won't this just add an extra step to your releases?

sanastos:
Ah, never mind, I see this change pulls from GCS and not from the github tag latest. LGTM.

PR #298, opened 2019-09-11
Added replicated cli builder

liiri:
https://github.com/replicatedhq/replicated


liiri:
@googlebot I consent.

sanastos:
Overall this PR looks fine to me, but can you modify the examples a bit?

For example, its best if at least one of the samples can be used as a sanity check by maintainers; other people might commit changes to this builder, and someone with less knowledge (like me) has to review and merge them, a simple test that passes helps give me confidence that the change should be accepted.

It seems like you need API keys to do anything useful with this builder, so maybe just print the version or help in the sanity check.

As for the other tests, please split them into something like `cleartext-cloudbuild.yaml` and `kms-cloudbuild.yaml`. Technically that file is currently invalid since it has two build steps with the same ID at the moment. You can leave the sanity check as `cloudbuild.yaml`.

Consider adding a quick README.md file in the directory explaining the contents.

PR #328, opened 2019-11-25
adds ko builder image

poy:
This adds a builder image for github.com/google/ko.

### DIFF
--- a/ko/Dockerfile
+++ b/ko/Dockerfile
@@ -9,10 +9,8 @@ COPY --from=0 /go/bin/ko /
 COPY ko.bash /builder/ko.bash
 
 # Install Go
-RUN curl https://dl.google.com/go/go1.13.4.linux-amd64.tar.gz -o go1.13.4.linux-amd64.tar.gz \
-  && tar -C /usr/local -xzf go1.13.4.linux-amd64.tar.gz \
-  && rm go1.13.4.linux-amd64.tar.gz
-ENV PATH="${PATH}:/usr/local/go/bin"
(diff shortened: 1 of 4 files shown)


ko/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+ENV GO111MODULE=on
+RUN CGO_ENABLED=0 GOOS=linux go get github.com/google/ko/cmd/ko
+
+FROM gcr.io/cloud-builders/kubectl
+
+COPY --from=0 /go/bin/ko /
+COPY ko.bash /builder/ko.bash
+
+# Install Go
```
ImJasonH: I wonder whether you can copy the Go binary out of the builder image... 🤔

poy: That's a cool idea. I'll try that out.

poy: Looks like that works

poy:
@ImJasonH I made the requested update, PTAL.

ko/cloudbuild.yaml
```diff
@@ -0,0 +1,5 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/$PROJECT_ID/ko', '.']
```
ImJasonH: Can you add a step that sanity checks that ko is available and installed?

```
- name: 'gcr.io/$PROJECT_ID/ko'
  args: ['version']
```

poy: Done

poy:
@ImJasonH I added the step, PTAL

PR #333, opened 2019-12-05
Enable versions in helm image tags

bery:
Add possibility to easily create, version and manage helm image versions using a single cloud build substitution.


sanastos:
LGTM, thanks for your contribution!

PR #336, opened 2019-12-10
ko: only authenticate if necessary

poy:
Only authenticate (via `/builder/kubectl.bash`) if the environment
variable `CLOUDSDK_CONTAINER_CLUSTER` is populated. This is helpful for
when the image is being used to do things other than applying (e.g.,
resolve).


leeonlee:
Thanks!

PR #337, opened 2019-12-11
Upgraded yarn-puppeteer dockerfile to use puppeteer 1.19.0

fernandoneira:
v1.19.0 plays nicely with the current major chrome version available on Alpine (77)


fernandoneira:
@googlebot I signed it!

PR #345, opened 2020-01-10
First version with new parameters to execute commands and create wind…

jmalvarezf-lmes:
Hi,

Here at Leroy Merlin Spain we need to compile some solutions that are Microsoft-Only and the only way we found it possible to accomplish the build with Google Cloud Build was using the windows remote cloud-build-step. As we found some difficulties for our deployment (network and regions basically) and we found that the method to copy workspace was very slow, we customized the windows-builder with success and we wanted to share the changes in case they are found useful.

Basically we added the following parameters:

- network
- bubnetwork
- region
- zone
- not-copy-workspace

We also made a very small change to allow the builder to get the password from an env var.

Hope it is useful. Feel free to ask for any change.

Thanks!

### DIFF
--- a/windows-builder/README.md
+++ b/windows-builder/README.md
@@ -51,6 +51,31 @@ steps:
           '--command', '<command goes here>' ]
 ```
 
+You can also provide the VPC, Subnetwork, Region, and Zone parameters to specify where to create the VM:
+
+```yaml
+steps:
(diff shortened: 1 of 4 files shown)


jmalvarezf-lmes:
@googlebot I signed it!

PR #368, opened 2020-02-21
Update cloudbuild.yaml

samirspatel:


samirspatel:
@googlebot I signed it!

PR #372, opened 2020-02-25
windows-builder: change access scope to `cloud-platform`

poy:
This follows the best practice recommended in the docs:

>A best practice is to set the full cloud-platform access scope on the
instance, then securely limit the service account's API access with
Cloud IAM roles.

https://cloud.google.com/compute/docs/access/service-accounts#accesscopesiam


LOZORD:
I was able to build this successfully. LGTM and thanks for your contribution!

PR #373, opened 2020-02-25
windows-builder: use latest Go

poy:
Previously, the version of Go was fixed to an old version. This commit
adds logic to find the latest version of Go instead.


poy:
@nof20 @LOZORD 

Let me do a little more work on this. I think I would like to change the API of the go-windows to where a user could build the docker image and then use it directly instead of having to use the docker CLI. For example, this would allow users to easily use this to test their Go code for example.

poy:
Actually, I'll make that a different PR for that. I'll make it a whole new container that can be built so that no one's API has a breaking change.

LOZORD:
This builds successfully for me! LGTM and thanks for your contribution!

PR #395, opened 2020-04-15
[terraform/packer] Update terraform to v0.12.26, packer to 1.6.0

xSAVIKx:
The aim of the PR is to make the latest `terraform` version the default one.


xSAVIKx:
@prabenzom I've just synced with the latest master and looks like updated the examples repo >_<
Is it OK or should I revert or modify it anyhow?

LOZORD:
I might be better to just create a new PR since this seems pretty old.

PR #412, opened 2020-05-16
doktl: Update cli tools

JMSwag:
Use an available kubectl binary.


JMSwag:
@googlebot I signed it!

PR #419, opened 2020-06-02
Bumped scala-sbt build to 1.3.12

aaronp:
Updated sbt version for scala-sbt builds


aaronp:
@googlebot I signed it!

PR #420, opened 2020-06-04
Improvements of slackbot

Sytten:
- Modernize the build process
- Add a new message when build starts
- Add a project variable to differentiate between builds
Screen Shot 2020-06-04 at 4 20 53 PM


Sytten:
@googlebot I signed it!

PR #436, opened 2020-07-29
Update protoc builder

xSAVIKx:
In this PR I have updated the `protoc` builder to use the latest v3.12.4 version of the protoc.
Also, I've cleaned up docs and switched the base image from `ubuntu:19.10` to smaller `debian:buster-slim`.

I was considering `alpine`, but its compiler is not compatible with protoc and I don't think the effort of making it work worth it.

### DIFF
--- a/protoc/Dockerfile
+++ b/protoc/Dockerfile
@@ -4,6 +4,8 @@ ARG VERS=3.12.4
 ARG ARCH=linux-x86_64
 
 RUN echo "Building protoc Cloud Builder ${VERS}-${ARCH}" && \
+    apk --update add --no-cache wget unzip && \
+    rm -rf /var/cache/apk/* && \
     wget "https://github.com/protocolbuffers/protobuf/releases/download/v${VERS}/protoc-${VERS}-${ARCH}.zip" && \
     unzip "protoc-${VERS}-${ARCH}.zip" -d protoc && \
     rm "protoc-${VERS}-${ARCH}.zip"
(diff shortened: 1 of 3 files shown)


protoc/cloudbuild.yaml
```diff
@@ -9,9 +9,11 @@ steps:
     - "--build-arg=VERS=${_VERS}"
     - "--build-arg=ARCH=${_ARCH}"
     - --tag=gcr.io/${PROJECT_ID}/protoc:${_VERS}-${_ARCH}
+    - --tag=gcr.io/${PROJECT_ID}/protoc:latest
```
LOZORD, line 12: Up to you: keep the same order here as below with `images`.

protoc/Dockerfile
```diff
@@ -1,17 +1,16 @@
-FROM ubuntu:19.10
+FROM debian:buster-slim
```
LOZORD, line 1: Up to you: When I ran `gcloud builds submit . `, I saw `debconf: delaying package configuration, since apt-utils is not installed` in the logs. Do we want to add this to the Dockerfile now that we're using Debian?

protoc/cloudbuild.yaml
xSAVIKx: Done.

protoc/Dockerfile
xSAVIKx: I don't think it worth doing. We're not really going to have any interactive installations where `apt-utils` could be useful.

LOZORD: Fine with me, thanks 👍

LOZORD:
Awesome, thanks again 😎 👍

PR #438, opened 2020-08-02
Windows builder bug fixes + improvements

dorzeidman:
Windows Builder:
Bug fixes:
1. If command exited with an error (ExitCode != 0) than the step will fail.
2. If Copy or Run command failed the instance will be deleted.
3. Wait operation also checking the Error property after operation is done.

Improvments:
1. Labels: added **labels** optional argument, labels will be added to new instance, similar to gcloud cli.
2. Machine Type: added **machineType** optional argument.
3. Command Timeout: added **commandTimeout** optional argument. default is still 5 minutes.
4. Copy Timeout: added **copyTimeout** optional argument. default is still 5 minutes.
5. Service Account: added **serviceAccount** operation argument. if **serviceAccount** is not an email, will add @$ProjectId.iam.gserviceaccount.com suffix


codrienne:
Thanks for the PR! I plan to review it (and probably merge, from a first glance) by the end of this week, hopefully.

PR #455, opened 2020-10-12
[envsubst] [kustomize] cloudbuild fixups

afirth:
## kustomize
- use the `LATEST_TAG` correctly during smoke test

## envsubst
- allow overriding the build directory, so that cloudbuild triggers can be created directly from a fork of this repository

P.S. if anyone is reading this for feature requests, _please_ consider allowing defaulting the build_dir to something else in the trigger yaml, because this is gross. But it works and most other things about cloudbuild are great, so thanks!

coauthored by: @Faffnir


Faffnir:
@googlebot I fixed it.

alygerharter (approved):

PR #466, opened 2020-12-01
Remote builder should wait for the instance to start accepting SSH connections

natarajaya:
Current version of `remote-builder` script spins up instance and then tries to scp workspace to it immediately.
This causes some builds to fail, because when `${GCLOUD} compute instances create` command finishes, instance is some times not yet ready to accept SSH connections:

```
+ trap cleanup EXIT
++ pwd
+ gcloud compute scp --compress --recurse /workspace cloud-user@builder-4c51fa5f-d6ee-4644-8634-1cf0cd8fcb11:/home/cloud-user/workspace/ --ssh-key-file=builder-key
ssh: connect to host 34.123.24.126 port 22: Connection refused
lost connection
ERROR: (gcloud.compute.scp) [/usr/bin/scp] exited with return code [1].
```

Added loop should ensure that instance is reachable over SSH before proceeding to SCP.

### DIFF
--- a/remote-builder/run-builder.sh
+++ b/remote-builder/run-builder.sh
@@ -14,6 +14,7 @@ INSTANCE_NAME=${INSTANCE_NAME:-builder-$(cat /proc/sys/kernel/random/uuid)}
 ZONE=${ZONE:-us-central1-f}
 INSTANCE_ARGS=${INSTANCE_ARGS:---preemptible}
 GCLOUD=${GCLOUD:-gcloud}
+RETRIES=10
 
 ${GCLOUD} config set compute/zone ${ZONE}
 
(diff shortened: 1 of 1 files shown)


natarajaya:
@googlebot I signed it!

remote-builder/run-builder.sh
```diff
@@ -31,6 +32,18 @@ ${GCLOUD} compute instances create \
        --metadata block-project-ssh-keys=TRUE \
        --metadata-from-file ssh-keys=ssh-keys
 
+READY_COMMAND="${GCLOUD} compute ssh --ssh-key-file=${KEYNAME} ${USERNAME}@${INSTANCE_NAME} -- printf 1"
```
LOZORD, line 35: Would it be better for this to be a function?

LOZORD, line 35: Paranoia: have this print 'OK' instead so we don't have to worry about stopping on an error code?

LOZORD (commented):
Thanks for your contribution! I have a few minor suggestions but otherwise this LGTM.

remote-builder/run-builder.sh
natarajaya: Makes sense.

natarajaya: I don't think it can cause any issues, but let's change it to be on the safe side.

natarajaya:
Hi @LOZORD, thanks for your review!
I addressed both of your comments.

LOZORD (approved):
This builder builds for me - LGTM ✅

PR #468, opened 2020-12-22
update helmfile and underlying helm to latest versions

naseemkullah:
https://github.com/roboll/helmfile/releases
https://github.com/helm/helm/releases

### DIFF
--- a/helmfile/cloudbuild.yaml
+++ b/helmfile/cloudbuild.yaml
@@ -5,9 +5,9 @@ steps:
         "build",
         "--tag=gcr.io/$PROJECT_ID/helmfile",
         "--build-arg",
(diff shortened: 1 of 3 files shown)


rimusz:
LGTM

helmfile/cloudbuild.yaml
```diff
@@ -5,9 +5,9 @@ steps:
         "build",
         "--tag=gcr.io/$PROJECT_ID/helmfile",
         "--build-arg",
-        "HELM_VERSION=v3.3.4",
+        "HELM_VERSION=v3.4.2",
```
leeonlee, line 8: Nit: these build args are redundant; they can be omitted if they're going to be the same values as the default ARG values.

PR #476, opened 2021-01-15
Refactor Kustomize dockerfile and bump default version

jmthvt:
Initially I just wanted to bump the default Kustomize version, but I also refactored the Dockerfile to take advantage of Docker multi-stage build.


kustomize/Dockerfile
```diff
@@ -1,25 +1,14 @@
...
-  apt-get remove --purge -y wget && \
-  apt-get --purge -y autoremove && \
-  apt-get clean && \
-  rm -rf /var/lib/apt/lists/*
 
-ENV PATH=/builder/kustomize:$PATH
+FROM gcr.io/cloud-builders/gcloud
 
+COPY --from=build kustomize /usr/bin/kustomize
+COPY kustomize.bash /builder/kustomize.bash
```
leeonlee, line 13: You likely will need to make these files executable.

jmthvt: True for the kustomize.bash file - nice catch.
The kustomize binary is already executable.

jmthvt: Made the file executable at source, resolving.

prabenzom (approved):

PR #491, opened 2021-02-22
updated readme for terraform and terragrunt

benking84:
Readme was a little confusing using terms that do not match to product names and lacking the getting started section, also fixed broken links


sanastos:
Thanks for your contribution! These docs had gotten pretty stale.

sanastos:
Actually, I see there is a conflict; can you resolve that so I can merge?

PR #494, opened 2021-03-18
Fix ko builds

josephlewis42:
ko requires golang 1.16 https://github.com/google/ko#build-and-install-from-source and the vendor directory to be used for correct dependency resolution which means the git repo has to be pulled in order to correctly build it without running into arbitrary breaking changes due to upstream libraries.

This PR also adds the ability to build a specific version of ko so users can choose to pin the version they use.


bendory (approved):
LGTM

PR #505, opened 2021-05-21
[jsonnet] install jsonnet after building

elisw93:
I often like to use Makefiles to help generate my `jsonnet` configuration, and I'm trying to make this Cloud Builder work with that kind of setup.

If `jsonnet` gets installed into `$PATH` after we build it, then I can set my entrypoint to `bash` or `make` and invoke `jsonnet` indirectly. That lets me do what I want without making any destructive changes.


elisw93:
@googlebot I signed it!

elisw93:
This has been sitting for a while, could I get an update?

PR #522, opened 2021-08-13
delete binary from repo.

perezd:
This isn't necessary/shouldn't be used.

### DIFF
--- a/windows-builder/.gitignore
+++ b/windows-builder/.gitignore
@@ -1,0 +1,1 @@
+builder/main


bendory (changes requested):
Would you mind adding a .gitignore as part of this PR so that this doesn't get uploaded by accident in the future?

Thanks for your PRs!

perezd:
I think that'll do it!

bendory (approved):

PR #523, opened 2021-08-17
upgrade helm and helmfile for helmfile builder and fix image

yashbhutwala:
- Upgrade helm and helmfile
- Fix `helmfile sync` example to point to correct image


yashbhutwala:
@googlebot I signed it!

prabenzom (approved):

Issue #537, opened 2021-10-13
Getting GCE error when trying to use the windows builder


nishant-shah-social:
## Affected builder image
gcr.io/cloud-builders-community/windows-builder

## Expected Behavior
Should be able to use the windows-builder image successfully in google cloud build

## Actual Behavior
Getting the below error message specific to GCE :-
`Starting GCE service in project playground-nishant
2021/10/12 06:35:00 GCE Instances insert call failed: googleapi: Error 400: Invalid value for field 'resource.tags.items[0]': ''. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)', invalid`
## Steps to Reproduce the Problem

1. Simply followed the steps mentioned here :- https://github.com/GoogleCloudPlatform/cloud-builders-community/tree/master/windows-builder

## Additional Info
I performed these steps from a mac book. Hope it should not be a problem

isarkis:
@perezd, is the --tags option mandatory? If so, we'll need to update the README.md.

perezd:
I didn't intend to make it required, does it appear to be? If so we can make it optional.

isarkis:
Yes, windows-builder throws an error without --tags argument.

perezd:
OK, that's a mistake. Send a PR to make it optional?

isarkis:
PTAL https://github.com/GoogleCloudPlatform/cloud-builders-community/pull/544

isarkis:
Can someone help close this issue, it's been fixed?

PR #544, opened 2021-11-15
Make --tags optional in windows-builder.

isarkis:
Do not use empty tags when creating GCE instance. This PR fixes issue #[537](https://github.com/GoogleCloudPlatform/cloud-builders-community/issues/537).

### DIFF
--- a/windows-builder/builder/builder/remote.go
+++ b/windows-builder/builder/builder/remote.go
@@ -223,6 +223,10 @@ func (bs *BuilderServer) GetLabelsMap() map[string]string {
 }
 
 func (bs *BuilderServer) GetTags() []string {
+  if *bs.Tags == "" {
+    return nil
+  }
+
 	var tags []string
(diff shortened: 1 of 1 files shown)


perezd:
LGTM

mwiczer:
Are you sure this fixes the issue? It looks to me (without testing it yet) that we need to remove the entry here when tags are empty: https://github.com/GoogleCloudPlatform/cloud-builders-community/blob/master/windows-builder/builder/builder/gce.go#L199-L201

perezd:
I could be wrong, but I thought that evaluated to an empty set of tags which is OK also.

mwiczer:
Oh yes, I see now. Without this change, there is a zeroth entry in "Items", causing
```
googleapi: Error 400: Invalid value for field 'resource.tags.items[0]': ''. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)', invalid
```
LGTM

Issue #545, opened 2021-11-15
[windows-builder] Handle WinRM communication over internal VPC network.


isarkis:
As it stands now windows-builder requires users to create a firewall rule to allow external access on TCP:5986.  This external firewall rule gives cloud workers ability to run arbitrary powershell commands on GCE instances via WinRM.  

Some users and orgs consider opening an external access to be a security threat.   To mitigate security concerns users can create a private worker pool in GCB and have it peered with an internal VPC network where GCE instances get created.  With such set up it's possible to have cloud worker communicate with GCE instance over an internal VPC network using internal IP addresses.

One way to do this is to add an option to have windows-builder reach GCE instance over an internal IP address (vs external IP).

PR #546, opened 2021-11-23
Add an option to use an internal ip address.

isarkis:
This PR addresses issue #545 by adding an option to have windows-builder communicate with GCE instance(s) over internal network.  This functionality can only work with a private worker pool peered with a private VPC network. Compared to the default behavior this should be a more secure option and avoids having users open an external access to their network via a firewall rule.

This PR has been tested with a test job that ran in a private worker pool over internal network.  It was also tested for regression by running a test job with default behavior.

### DIFF
--- a/windows-builder/builder/main.go
+++ b/windows-builder/builder/main.go
@@ -33,6 +33,8 @@ var (
 	copyTimeout      = flag.Int("copyTimeout", 5, "The workspace copy timeout in minutes")
 	serviceAccount   = flag.String("serviceAccount", "default", "The service account to use when creating the Windows server")
 	tags             = flag.String("tags", "", "List of strings eparated by comma to add when creating the Windows server")
+	useInternalNet   = flag.Bool("use-internal-network", false, "Communicate with Windows server over the internal network")
+	createExternalIP = flag.Bool("create-external-ip", false, "Create an external IP address when using internal network")
 )
 
 func main() {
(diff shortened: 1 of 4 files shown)


windows-builder/builder/main.go
```diff
@@ -65,6 +66,7 @@ func main() {
 			DiskType:       diskType,
 			ServiceAccount: serviceAccount,
 			Tags: 					tags,
+			InternalNet: useInternalNet,
```
agvdndor, line 69: I believe this should be  `UseInternalNet: useInternalNet,`

agvdndor:
Thanks for the great work! I just tried it out and there's still an issue with the creation of the instance. With the current way of defining an instance in `gce.go`, an external IP is still created: 
```
NetworkInterfaces: []*compute.NetworkInterface{
	&compute.NetworkInterface{
		AccessConfigs: []*compute.AccessConfig{
			&compute.AccessConfig{
				Type: "ONE_TO_ONE_NAT",
				Name: "External NAT",
			},
		},
		Network:    prefix + s.projectID + "/global/networks/" + *bs.VPC,
		Subnetwork: prefix + s.projectID + "/regions/" + *bs.Region + "/subnetworks/" + *bs.Subnet,
	},
},
```
This is solved by omitting the AccessConfigs altogether when the --use-internal-network flag is passed i.e.:
```
NetworkInterfaces: []*compute.NetworkInterface{
	&compute.NetworkInterface{
		Network:    prefix + s.projectID + "/global/networks/" + *bs.VPC,
		Subnetwork: prefix + s.projectID + "/regions/" + *bs.Region + "/subnetworks/" + *bs.Subnet,
	},
},
```
Works as expected after that.

windows-builder/builder/main.go
isarkis: fixed.

isarkis:
@agvdndor , I am leaning towards letting windows-builder create an external address for GCE instance.  I often times find myself need to RDP to windows GCE instance which requires an external IP address.  I am sure users can find it handy.  Also, without external IP address GCE instance won't be able to access GCS buckets.  A workaround would be to enable [Private Google Access](https://cloud.google.com/vpc/docs/configure-private-google-access) on the subnet used by a private worker pool.  Let me know what you think.

agvdndor:
@isarkis Fair point. However organizations/projects that have external IP addresses disabled as a policy will be unable to use this without modifications. Perhaps an extra input argument could be useful. I was surprised that it still created the external IP. It is really your call though.

isarkis:
I went ahead and added --create-external-ip option which is false by default.

agvdndor (approved):

isarkis:
Can someone help submit this PR, I don't have permission.

PR #547, opened 2021-11-24
Changes to the Terraform with CloudBuild at scale sample

AlexBulankou:
-  extracting common module
 -  generating per-environment configurations to follow best practices of storing per-environment configs.


terraform/examples/infra_at_scale/multi-build.sh
```diff
@@ -183,10 +183,10 @@ for i in "${!regions[@]}"; do
     gsutil mb -l $REGION gs://$BUCKET_NAME
     printf "Running in region %s, zone %s\n" "${REGION}" "${ZONE}"
     (gcloud builds submit  \
-        --worker-pool=projects/${PROJECT_ID}/locations/${WORKER_POOL_REGION}/workerPools/${WORKER_POOL_ID} . \
-        --timeout=1200s \
-        --config=cloudbuild.yaml \
-        --region=$WORKER_POOL_REGION \
-        --substitutions=_BUCKET=$BUCKET_NAME,_REGION=$REGION,_ZONE=$ZONE &)
+            --worker-pool=projects/${PROJECT_ID}/locations/${WORKER_POOL_REGION}/workerPools/${WORKER_POOL_ID} . \
```
morgante, line 186: Instead of repeating the region list here, you should have this script loop through the `environments` directory and deploy each zone found. This way you can add/remove zones just be deleting them from environments.

terraform/examples/infra_at_scale/generate_environments/generate.sh
```diff
@@ -0,0 +1,189 @@
...
+    us-west4-b
+    us-west4-c
+)
+
+for i in "${!regions[@]}"; do
+    REGION=${regions[i]}
+    ZONE=${zones[i]}
+    printf "Preparing environment in region=%s, zone=%s\n" "${REGION}" "${ZONE}"
+   
+    mkdir -p ./environments/$REGION/$ZONE
```
morgante, line 184: Not sure about nesting this by region. I might just make it a flat structure `environments/$ZONE`.

PR #571, opened 2022-04-19
Update cloudbuild.yaml for docker-compose

sheenpythian:
updating version to current and adding the "v" so that people know it is supposed to be there.


sheenpythian:
just killing this as it is apparently not going anywhere.

PR #582, opened 2022-06-23
Upgrade go version for ko image to 1.17

guangyw:
Installation of KO also changed to use `go install` from go1.16+


kkr16:
@guangyw can you please bump `golang` image to 1.18+?
builds of ko version >= 12.0 require golang >= 1.18

guangyw:
> @guangyw can you please bump `golang` image to 1.18+? builds of ko version >= 12.0 require golang >= 1.18

@kkr16 done

PR #581, opened 2022-06-22
Update Packer to release version 1.8.2

ejv:
This PR does the following:

- Updates Packer to release version `1.8.2`
- Updates Alpine to release version `3.16`.
- Updates the `source_image_family` in the example build.

Note, the previous Packer release version `1.8.1` included fixes to address issues with legacy SSH key algorithms that fundamentally arose due to upstream changes in [OpenSSH](https://www.openssh.com/txt/release-8.8). This typically manifested in the form of build failures due to Packer authentication errors (handshake failed) on contemporary OS base images. Client authentication support for algorithms based on SHA-2 for use with existing RSA keys is now implemented, which will mitigate this particular corner case.

PR #583, opened 2022-07-19
Adopt latest Packer release

tpdownes:
The current build for Packer installs Packer 1.7.8. Packer 1.7.9 (Nov 2021) supports using variables in the build name and it is useful in several contexts. This PR bumps to the present release of Packer, 1.8.2.


tpdownes:
This is a duplicate of #581. Please review that!

PR #591, opened 2022-11-05
Update packer example template

sdarwin:
When running packer to generate an image, the user will likely consider changing the template from ubuntu-2104 to ubuntu-2204, since that's a long term support version and it's newer.  After making that change, packer will fail.  The reason is not obvious. After logging into the build machine it says `userauth_pubkey: key type ssh-rsa not in PubkeyAcceptedAlgorithms `  The solution is to add `temporary_key_pair_type`.

### DIFF
--- a/packer/examples/gce/build.pkr.hcl
+++ b/packer/examples/gce/build.pkr.hcl
@@ -12,12 +12,13 @@ variable "builder_sa" {
 
 source "googlecompute" "test-image" {
   project_id                  = var.project_id
-  source_image_family         = "ubuntu-2104"
+  source_image_family         = "ubuntu-2204-lts"
   zone                        = var.zone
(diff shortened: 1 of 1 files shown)


ejv:
As of Packer release version [1.8.1](https://github.com/hashicorp/packer/releases/tag/v1.8.1), it is no longer necessary to declare `temporary_key_pair_type` in the build configuration (users will want to ensure they are using a contemporary version of Packer).

I've included the change to the `source_image_family` in https://github.com/GoogleCloudPlatform/cloud-builders-community/pull/604 (updating to Packer release version [1.8.5](https://github.com/hashicorp/packer/releases/tag/v1.8.5)).

I recommend closing this PR. Thanks for your contribution.

PR #608, opened 2023-01-31
Add Poetry build image

gilsondev:
This PR creates a new cloud build image with the [`poetry`](https://python-poetry.org/) command installed. The goal of this image is offer options to run some steps, for example _build_ and _publish_ python packages to package management like Pypi or Nexus.


mshamma:
I'm interested in using this builder if approved.  Any updates?

PR #617, opened 2023-05-05
feat(packer): Update Packer default to 1.8.7

ianchesal:
Update to the very latest stable packer release. Bump the version in the readme while we're there.


ianchesal:
Sigh. Seemed like an easy add. A year later and it's still pending. Giving up.

PR #622, opened 2023-06-05
Android gradle 8.0.0 support

routhiervincent:
The current android cloud builder doesn't support the Android Gradle plugin 8.0.0 because it use google javac builder which is based on Java 8. The new plugin require Java 17. 

The following commit make it work with gradle 8.0.0.


itok01:
I support this Pull Request.

Currently, the project created in Android Studio uses Gradle 8, so it cannot be built with the current Dockerfile-base.

haroonc (approved):

PR #705, opened 2026-03-17
Update gcloud builder to accept versions ARG for packer and terraform

jrossthomson:
Running some of the gcluster modules requires terraform to be at v1.12.2. This causes failures when doing a "gcluster deploy" command for certain version locked Cluster Toolkit modules.


jrossthomson:
@vyolla PTAL!

PR #715, opened 2026-06-07
Allow to use recent Golang and ko versions

balamut-google:
Use new version of Golang and parametrize the version number.
Also change the build command to work with new Go toolchain and recent ko repository structure.


nstandif:
+1

